PR #33, opened 2022-03-16
Adapting all workflows.

Narsil:
Updating the workflows so hopefully they work again.

This was done with sed magic, so no files should have been forgotten.


osanseviero (approved):
Looks good!

PR #40, opened 2022-05-04
Updating flair to `0.11.1`

Narsil:
Rebuilt the old version, but since it's using a non pinned version of `flair` it can cause issues (Lack of Etag seems to be linked to `huggingface_hub` not flair though, but I have seen that there could be some small modifications needed too.? )

### DIFF
--- a/.github/workflows/python-api-flair.yaml
+++ b/.github/workflows/python-api-flair.yaml
@@ -1,0 +1,26 @@
+name: flair-docker
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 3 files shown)


stefan-it:
Many thanks  @Narsil ! I tested it locally and it is working with recent Flair release!

osanseviero (approved):
It might be a good idea to add a GitHub Actions workflow to test this in https://github.com/huggingface/api-inference-community/tree/main/.github/workflows to make sure things run fine.

I'm approving to avoid blocking!

PR #44, opened 2022-05-05
Enable the override to take place (API needs to bypass WAF).

Narsil:


osanseviero (approved):
Thank you for the fix!! :hugs:

PR #45, opened 2022-05-06
Update all docker images to use `huggingface_hub==0.5.1` to have `HF_ENDPOINT` override available.

Narsil:


osanseviero (approved):
Looks good! THank you :fire: None of the libraries pin `huggingface_hub` in their versions, so this should work well!

PR #50, opened 2022-05-12
Remove unused pipeline files

### LINKED ISSUE #47: Remove unimplemented pipelines across the different integrations
**Is your feature request related to a problem? Please describe.**
Is there any reason why we have a ton of unused pipeline files across the different integrations ([example](https://github.com/huggingface/api-inference-community/blob/main/docker_images/asteroid/app/pipelines/image_classification.py))? I think it would be a lot easier to both navigate the repo and quickly see which integrations have which pipelines enabled if we only included the files that are actually being used. 

**Describe the solution you'd like**
- [ ] Remove unused pipeline files from the different integrations.
- [ ] Update guidance in contributing docs if need be to reflect the fact we shouldn't be copying over all the pipeline template files.

WDYT?

nateraw:
This PR resolves #47 

- Removed all unused pipeline files
- Removed all unused pipeline tests
- Added to README.md to explain you should clean up the unused pipeline files


osanseviero (approved):
Thanks for this PR!! :fire: :fire: this is so so so much better now

julien-c:
yes i 100% agree! way cleaner

Narsil:
Screenshot from 2022-05-12 09-15-04

The mark of an insanely good PR !

LGTM

Narsil:
Merging, speechbrain failure is linked to bad file caching (they use torch hub downloading iirc).

We can ignore, that's what is in prod.

Narsil (approved):

TParcollet:
I saw speechbrain so I jump in. Anything we should do with respect to this ?

Narsil:
> I saw speechbrain so I jump in. Anything we should do with respect to this ?

Hi, thanks for the help !  This is not a critical issue by any means.

The Docker image just needs information on where the cache of the model file will live.
By default for `huggingface_hub` library it will use HUGGINGFACE_HUB_CACHE: https://github.com/huggingface/api-inference-community/blob/main/docker_images/speechbrain/Dockerfile#L20

However somehow I think some files are cache missed during that step, aren't you using also torch hub for some files ? (Sorry I am not sure anymore). If yes, is there a way we could override this setting (to point it to a shared disk so subsequent loads are faster).

If not, I need to understand why it's redownloading the model, also please not that these tests are a bit crude and only test load times to see if the cache is hit, so maybe we're just testing a model that is a bit too big ?

TParcollet:
We are not using anything else and we only create symlinks towards these I believe. Here is the only stuff we do with respect to HF when using a pretrained model from the hub: https://github.com/speechbrain/speechbrain/blob/b000a29cb8f2b1a93bdad2916717579802c22b8d/speechbrain/pretrained/fetching.py#L110

PR #53, opened 2022-05-20
`--model-id=XX/XXX` becomes `XX/XXX` (simpler CLI).

Narsil:
--model-id=XX/XXX becomes XX/XXX (simpler CLI).

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -132,7 +132,7 @@ def resolve(model_id: str) -> [str, str]:
 
 
 def resolve_task_framework(args):
-    model_id = args.model_id
+    model_id = args.model
     task = args.task
     framework = args.framework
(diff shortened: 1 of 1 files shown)


manage.py
```diff
@@ -199,9 +199,8 @@ def main():
         "start", help="Start a local version of a model inference"
     )
     parser_start.add_argument(
-        "--model-id",
+        "model",
         type=str,
-        required=True,
```
osanseviero, line 204: shouldn't this be required?

osanseviero (approved):
Looks good!

manage.py
Narsil: No, positional arguments are necessarily required (it actually raises an exception if that flag is on :)

PR #59, opened 2022-06-01
ENH Replacing Structured Data Classification with Tabular Regression and Tabular Classification

merveenoyan:
This PR solves [#137 in hub-docs](https://github.com/huggingface/hub-docs/issues/137).

The renames done: 

Pipeline related classes are replaced with following:
- TabularDataInputChecks
- TabularDataPipeline 

Above one are used for both `tabular-classification` and `tabular-regression` tasks.

For specific libraries and their tests, there's only tests for `tabular-classification` as it was what was solved. Let me know if you want me to duplicate this to have tests for regression, they do input and return the same thing so I didn't duplicate it for the sake of simplicity.


adrinjalali (commented):
The `tabular-regression` doesn't seem to be tested.

merveenoyan:
@adrinjalali if you want I can test that. Since this and `tabular-classification` refer to the same thing (by means of pipeline) atm I didn't want to duplicate code.

docker_images/generic/tests/test_api_tabular_regression.py
```diff
@@ -0,0 +1,96 @@
...
+        {"model_id": model_id}
+        for model_id in TESTABLE_MODELS["tabular-regression"]
+    ]
+)
+class TabularRegressionTestCase(TestCase):
+    def setUp(self):
+        self.old_model_id = os.getenv("MODEL_ID")
+        self.old_task = os.getenv("TASK")
+        os.environ["MODEL_ID"] = self.model_id
+        os.environ["TASK"] = "tabular-regression"
```
adrinjalali, line 25: are these env variables required inside some of the methods? Ideally you could just store them somewhere else or use original values if they're set. But I'm happy at least having some tests :)

osanseviero: Yes, they are set in the docker image when spawned. @Narsil can give more details, but the answer is they are used indeed

Narsil (commented):
Hi Thank you for tacking this  !

In order to get this live, we will need to first publish a new version of `api-inference-community`. 
To be perfectly clean this would require 2 PRs (1 for adding things in the lib, 1 for updating the community images).

We can fasttrack this if you can just update the version number in `setup.py` + adress the other comments.

PR #60, opened 2022-06-07
Add SpeechBrain TTS

osanseviero:
[Still draft]


osanseviero:
The unit test passes. The docker test is failing but in another part (ASR rather than TTS)

docker_images/speechbrain/requirements.txt
```diff
@@ -3,3 +3,4 @@ api-inference-community==0.0.23
 huggingface_hub==0.5.1
 speechbrain==0.5.11
```
Narsil, line 4: We need to remove this if we use a git version, no ?

Narsil (changes requested):
LGTM

PR #64, opened 2022-06-08
Add timm to requirements for FastAI

NimaBoscarino:
This adds `timm==0.5.4` to the default requirements for FastAI environments, as per the discussion on Slack: https://huggingface.slack.com/archives/C01BWJU0YKW/p1654712381652479?thread_ts=1654708826.292169&cid=C01BWJU0YKW

### DIFF
--- a/docker_images/fastai/requirements.txt
+++ b/docker_images/fastai/requirements.txt
@@ -1,3 +1,4 @@
 starlette==0.14.2
 api-inference-community==0.0.23
 huggingface_hub[fastai]==0.6.0
+timm==0.5.4


nateraw (approved):
Based on convo in slack, this LGTM

omarespejel (approved):

omarespejel:
LGTM! This would be very helpful now that fastai supports `timm` models 🤗.

osanseviero (commented):
Doesn't this then assume all uploaded `fastai` models use exactly the same version of `timm`? Maybe it would be good to add a test with a model that was not working before and is working now at https://github.com/huggingface/api-inference-community/blob/main/tests/test_dockers.py

Narsil:
I think @osanseviero is spot on, may you add a test of a real model that should now work.

I took the liberty of doing it if you don't mind. (Just wanted to check if it was working and since the work was already done...)

PR #67, opened 2022-06-29
ENH create a new env for each sklearn model

adrinjalali:
This PR makes the inference side to create an environment for each sklearn model and then spawn a new process for the server.

There's quite a bit which needs to be done here, but before I jump into those, wanted to hear your initial thoughts, hence the early dirty hackish PR.

TODO:
- [ ] add documentation
- [x] add example usage with example existing repo
- [x] add tests
- [x] fix the docker file with the new requirements

I'd be interested in your thoughts @Narsil @osanseviero 

__EDIT (30-06-2022)__
This PR can be tested locally as:

```bash
./manage.py start --framework sklearn --task=tabular-classification scikit-learn/iris-demo
```

and sending the request: 

```bash
curl -X POST -d @docker_images/sklearn/tests/samples/iris.json http://localhost:8000
```


osanseviero (commented):
Conceptually LGTM from my side, this would be very cool! Happy to do a proper review once it's not a draft

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,18 +1,37 @@
...
+        cached_folder = snapshot_download(repo_id=model_id)
+        print(cached_folder)
+        try:
+            with open(Path(cached_folder) / "config.json") as f:
+                # this is the default path for configuration of a scikit-learn
+                # project. If the project is created using `skops`, it should have
+                # this file.
+                config = json.load(f)
+            print(config)
+            model_file = config["sklearn"]["model"]["file"]
```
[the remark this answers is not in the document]
adrinjalali: It's not adding complexity really. We're just not forcing people with a specific name, and there doesn't seem to be anything really depending on the hard coded name. I'm not sure why the model should ideally always have the same name. What are we trying to infer from the name that makes it significant?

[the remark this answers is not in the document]
julien-c: >  the model should ideally always have the same name 

for transformers for instance it was IMO important for adoption to have a principle of least surprise/consistency/Developer experience.

Is there any advantage to be able to pick an arbitrary filename?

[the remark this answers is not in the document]
adrinjalali: To me, developer experience is why I want it to be flexible. scikit-learn is old and people have very different behaviors and habits when it comes to saving their models, and I don't see why we should enforce one over another.

Also, some people might save it with joblib and some with pickle, they're not even the same thing necessarily.

merveenoyan:
I'll take over whatever you would like a hand on 🙂 @adrinjalali

docker_images/sklearn/app/pipelines/tabular_classification.py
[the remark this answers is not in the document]
merveenoyan: I agree with @adrinjalali here, unlike transformers, there's not much consistency across the developer behavior of scikit-learn since everyone learns it from different sources. Some people call their models model, some call it clf, some call it pipeline, they also serialize differently.

adrinjalali:
There's something I'm not sure how to solve:

The tests in `docker_images/sklearn/tests/test_api_tabular_classification.py` run a temporary server kinda thing and send a request to get the answer.

But the new system creates a new environment for each model, otherwise there will be a warning that the model is loaded with the wrong version of sklearn (we can also ignore that warning, but there are then no guarantees for it to keep working, and it would depend on the version of sklearn installed on the developer's machine/CI). I'm not sure how to solve it.

I'll work on the docker image to get sth working on that side of things.

adrinjalali:
The docs say:

> 1. Test your creation within a docker
> 
> ```python
> ./manage.py docker --model-id MY_MODEL
> ```
> 
> should work and responds on port 8000. `curl -X POST -d "test" http://localhost:8000` for instance if 
> the pipeline deals with simple text.

But the dockerfile doesn't really state anything to be executed. Not sure how the docker images are tested then. Ideas?

osanseviero:
FYI you can also add integration tests that run Docker at https://github.com/huggingface/api-inference-community/blob/main/tests/test_dockers.py

adrinjalali:
I'm not sure how to move forward here since I think I don't have all the necessary information.

The docker images in this repo seem to be missing any CMD or an ENTRYPOINT, and the `manage.py` file which I'm modifying here to add the logic isn't even copied into the docker files.

@Narsil any pointers?

adrinjalali:
I'm not sure how to move forward here since I think I don't have all the necessary information.

The docker images in this repo seem to be missing any CMD or an ENTRYPOINT, and the `manage.py` file which I'm modifying here to add the logic isn't even copied into the docker files.

@Narsil any pointers?

docker_images/sklearn/app/pipelines/tabular_classification.py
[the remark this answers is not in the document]
osanseviero: @merveenoyan you could say the same for TF in which some users use h5 and others SavedModel, but we decided to choose consistency :D 

The official docs suggest joblib or pickle, so I think supporting both formats is ok, but I would at least try to be consistent with the model filename. I don't see an advantage of letting any filename. Also the way they name the variable is totally ok and up to the users (users use `pipe`, `model`, `generator`, etc with `transformers` stuff), it's more about how the model file is named.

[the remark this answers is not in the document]
adrinjalali: Imagine this PR is coming from a developer of a package from the community, and they want to give their users this flexibility. They're implementing it, and it's not adding much complexity on our side, I don't see why we'd try to enforce HF ideas on them.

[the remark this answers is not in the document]
merveenoyan: @osanseviero Keras models are not serialized through .h5 for a while because google sort of bought it and provided better option to have SavedModel to leverage TFX. they literally only have two options and one of them is almost deprecated 😅 but people still want to use .h5 in Hub (though we do not support it) given we've seen PRs or issues on `huggingface_hub` repository and discord. If it doesn't add complexity on our side, I'm not in favor of forcing developer behavior in a monocultural way. We can call it X in our docs or encourage the use of X but not enforce it.

manage.py
```diff
@@ -154,12 +155,30 @@ def start(args):
...
+        # This means the task has a separate step to create an environment
+        # before running the model. We should therefore first call it, and then
+        # run the model in another process with the newly created environment.
+        print(1)
+        subprocess.run(["python", str(env_creator)], env=os.environ)
+        print(2)
+        with cd(local_path):
+            subprocess.run(["bash", str(app_runner)], env=os.environ)
+        print(3)
+    else:
```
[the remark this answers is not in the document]
Narsil: You don't HAVE to use `starlette` (which we use for http) so you can definitely change the command, but it doesn't seem to be what you are doing here, right ?

Narsil (commented):
Looks very good.

IMO,  the only thing is to add the `env_setup.py` as a first command in `CMD`.

You could just edit `start.sh` (defined in tiangolo/xxx) and do something like 

`python env_setup.py && whatever_was_there_before `.

Wdyt ?

docker_images/sklearn/app/pipelines/tabular_classification.py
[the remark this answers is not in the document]
osanseviero: Up to you :) you know better the ecosystem but please make sure to consider the trade-offs.

> I don't see why we'd try to enforce HF ideas on them.

Consistency has played a big role in adoption in the past. With the proposal, we might  see **many** beginners confused why their model does not work/load properly and the reason being that they don't have a config file with a sklearn key with a model key with a file key, which adds complexity and potential sources of errors for users. If we're confident `skops` will handle all of this very nicely then it might not be such a problem though :)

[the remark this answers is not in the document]
adrinjalali: I very much agree with all those arguments. I'm a firm believer of "simple things should be easy, complicated things should be possible". That's why I haven't removed the default name and I think a repo with the default name should still work. Although we still need to handle dependencies. But I completely agree with what you're trying to achieve.

docker_images/sklearn/Dockerfile
```diff
@@ -1,29 +1,40 @@
-FROM tiangolo/uvicorn-gunicorn:python3.8
-LABEL maintainer="Omar Sanseviero omar@huggingface.co"
+FROM mambaorg/micromamba
```
osanseviero, line 1: Are there any Python requirements we could pre-install to make the loading as fast as possible? (i know sklearn is dynamic, but more generally I was wondering about other deps)

docker_images/sklearn/app/pipelines/__init__.py
```diff
@@ -1,3 +1,3 @@
 from app.pipelines.base import Pipeline, PipelineException  # isort:skip
 
-from app.pipelines.tabular_data_pipeline import TabularDataPipeline
+from app.pipelines.tabular_classification import TabularDataPipeline
```
osanseviero, line 3: Maybe let's rename the class as well

```suggestion
from app.pipelines.tabular_classification import TabularClassificationPipeline
```

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,18 +1,34 @@
...
+                # project. If the project is created using `skops`, it should have
+                # this file.
+                config = json.load(f)
+            model_file = config["sklearn"]["model"]["file"]
+        except Exception as e:
+            print(e)
+            # If for whatever reason we fail to detect requirements of the project,
+            # we install the latest scikit-learn.
+            # TODO: we should log this, or warn or something.
+            model_file = DEFAULT_FILENAME
```
osanseviero, line 29: For better error handling/messaging I think instead of a `try`/`Exception`, checking for existence of file and respective config keys might be better. Up to you!

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,18 +1,34 @@
...
 
 DEFAULT_FILENAME = "sklearn_model.joblib"
 
 
 class TabularDataPipeline(Pipeline):
     def __init__(self, model_id: str):
-        self.model = joblib.load(
-            open(cached_download(hf_hub_url(model_id, DEFAULT_FILENAME)), "rb")
-        )
+        cached_folder = snapshot_download(repo_id=model_id)
```
osanseviero, line 16: Why not use `skops` instead?

docker_images/sklearn/run_app.sh
```diff
@@ -0,0 +1,53 @@
...
+# script.
+response="$(curl https://huggingface.co/$MODEL_ID/raw/main/config.json -f --output /tmp/config.json)" || response=$?
+if [ -z $response ]; then
+    requirements=$(get_requirements)
+else
+    # if the curl command is not successful, we use a default set of
+    # dependencies, and use the latest scikit-learn version. This is to allow
+    # users for a basic usage if they haven't put the config.json file in their
+    # repository.
+    requirements="pandas uvicorn gunicorn scikit-learn"
```
osanseviero, line 39: Nice!

docker_images/sklearn/tests/test_api.py
```diff
@@ -5,10 +5,17 @@
...
-# Tests do not check the actual values of the model output, so small dummy
-# models are recommended for faster tests.
-TESTABLE_MODELS: Dict[str, str] = {"tabular-classification": "julien-c/wine-quality"}
+# Must contain at least one example of each implemented pipeline Tests do not
+# check the actual values of the model output, so small dummy models are
+# recommended for faster tests. Note: make sure all models require
+# scikit-learn=1.0.2, and if you change that, make sure README.md of sklearn
+# folder is also updated.
+TESTABLE_MODELS: Dict[str, str] = {
+    "tabular-classification": {
```
osanseviero, line 14: It might make sense to have a couple extra tests for some of the other conditions in the pipeline
* What if file in config.json is invalid (e.g. no sklearn key)
* Missing config.json

The following test structure might be useful here https://github.com/huggingface/api-inference-community/blob/main/docker_images/speechbrain/tests/test_api.py#L11

osanseviero (commented):
This is super cool! Very excited about this 🚀 

I left some high level questions, but overall this is super nice!

docker_images/sklearn/Dockerfile
adrinjalali: we're kinda not even fixing the python version here. For an older version of sklearn which doesn't support a new python release, we'll have the latest python that it supports, and for the new ones the latest python release. And once python itself becomes not fixed, I'm not sure what we can do about the rest lol

But we can do a local channel deployment at some point on our infra to reduce traffic and increase speed. A basic sklearn env at this point does about 140MB of download. We could also have a proxy and cache all of those.

docker_images/sklearn/app/pipelines/tabular_classification.py
adrinjalali: I had two reasons in mind:

- skops would be an extra dependency, which we don't need if we're only downloading the model
- we don't care about caching here, if anything, it's a good thing and we don't care about where the model is downloaded.

So at the end it didn't seem like we really need the dependency. If in the future more features come in skops which we might need here, then we can change that.

docker_images/sklearn/app/pipelines/tabular_classification.py
adrinjalali: For the end user it doesn't matter why it fails, we're not showing them any warnings ATM. I'm not sure how to propagate warnings or messages here, that's why I put the TODO above this line. When we do that, then we can catch different exception classes and do the right warning (doesn't have to be an if/else statement for this purpose). Also: https://docs.python.org/3.10/glossary.html#term-eafp

docker_images/sklearn/tests/test_api.py
adrinjalali: Yep, will get a space for those repos and add the tests

(cc @merveenoyan this is where we need those repositories we just talked about offline)

docker_images/sklearn/app/pipelines/tabular_classification.py
adrinjalali: Oh I mean quite the other way around. In skops I wanted to make sure the cache is deleted and there is only one copy locally, and that it's on the path that the user decides. `snapshot_download` does caching and everything, so we're fine for here.

adrinjalali: It's already there ;)

osanseviero: Great! Sorry, I just saw the diff. Thanks :+1:

Narsil (approved):
LGTM.

Don't think there's any roadblock left right ?

I just happened to see maybe an issue:


```diff
diff --git a/docker_images/sklearn/app/main.py b/docker_images/sklearn/app/main.py
index 484c49a..adc4dc7 100644
--- a/docker_images/sklearn/app/main.py
+++ b/docker_images/sklearn/app/main.py
@@ -4,7 +4,7 @@ import os
 from typing import Dict, Type

 from api_inference_community.routes import pipeline_route, status_ok
-from app.pipelines import Pipeline, TabularDataPipeline
+from app.pipelines import Pipeline, TabularClassificationPipeline
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
 from starlette.middleware.gzip import GZipMiddleware
@@ -34,7 +34,7 @@ logger = logging.getLogger(__name__)
 # directories. Implement directly within the directories.
 ALLOWED_TASKS: Dict[str, Type[Pipeline]] = {
     # IMPLEMENT_THIS: Add your implemented tasks here!
-    "tabular-classification": TabularDataPipeline
+    "tabular-classification": TabularClassificationPipeline
 }
 ```
 
 Thanks for doing this. Happy to fastrack this whenever you feel it's ready !

File: docker_images/sklearn/app/pipelines/tabular_classification.py [hunk 9]
```diff
@@ -27,4 +43,7 @@ def __call__(
         """
         column_values = list(inputs["data"].values())
         rows = list(zip(*column_values))
-        return self.model.predict(rows).tolist()
+        # TODO: we should do sth with the warnings.
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            return self.model.predict(rows).tolist()
```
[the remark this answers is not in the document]
adrinjalali: If we don't catch these warnings here, there will always be a 500 response if the user has used a pandas DataFrame to train their model. The new versions of `sklearn` store the feature names used during `fit` and expect the same when `predict` is called, otherwise raise a warning. I don't want to fix those issues here in this PR, rather get something running and then fix the remaining issues in the coming PRs if that's okay with you.

adrinjalali:
> LGTM.
> 
> Don't think there's any roadblock left right ?
> 
> I just happened to see maybe an issue:
> 
> ```diff
> diff --git a/docker_images/sklearn/app/main.py b/docker_images/sklearn/app/main.py
> index 484c49a..adc4dc7 100644
> --- a/docker_images/sklearn/app/main.py
> +++ b/docker_images/sklearn/app/main.py
> @@ -4,7 +4,7 @@ import os
>  from typing import Dict, Type
> 
>  from api_inference_community.routes import pipeline_route, status_ok
> -from app.pipelines import Pipeline, TabularDataPipeline
> +from app.pipelines import Pipeline, TabularClassificationPipeline
>  from starlette.applications import Starlette
>  from starlette.middleware import Middleware
>  from starlette.middleware.gzip import GZipMiddleware
> @@ -34,7 +34,7 @@ logger = logging.getLogger(__name__)
>  # directories. Implement directly within the directories.
>  ALLOWED_TASKS: Dict[str, Type[Pipeline]] = {
>      # IMPLEMENT_THIS: Add your implemented tasks here!
> -    "tabular-classification": TabularDataPipeline
> +    "tabular-classification": TabularClassificationPipeline
>  }
> ```
> 
> Thanks for doing this. Happy to fastrack this whenever you feel it's ready !

I've fixed the one in sklearn folder, but we have a few other instances of `TabularDataPipeline` which I'm not sure if they need to be changed, including in `fastai`.

PR #69, opened 2022-07-05
Proof of concept for diffusers API

osanseviero:
This would not work well in current state as inference would be super slow and the loading would take significant time due to very large models. For launch we'll focus in spaces


osanseviero:
Closing in favour of #76

PR #76, opened 2022-07-27
Adding support for `diffusers`

Narsil:


docker_images/diffusers/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM tiangolo/uvicorn-gunicorn:python3.8
+LABEL maintainer="Nicolas Patry <nicolas@huggingface.co>"
+
+# Add any system dependency here
+# RUN apt-get update -y && apt-get install libXXX -y
+
+RUN pip install --no-cache-dir torch==1.12.0+cpu torchvision==0.13.0+cpu torchaudio==0.12.0 --extra-index-url https://download.pytorch.org/whl/cpu
```
osanseviero, line 7: Why is this here rather than in `requirements.txt`?

docker_images/diffusers/app/pipelines/text_to_image.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, model_id: str):
+
+        kwargs = (
+            {"safety_checker": None}
+            if model_id.startswith("hf-internal-testing/")
+            else {}
+        )
+        self.ldm = DiffusionPipeline.from_pretrained(
+            model_id, use_auth_token=os.getenv("HF_API_TOKEN"), **kwargs
+        )
```
osanseviero, line 22: Should we set a seed or should we have no seed?

docker_images/diffusers/tests/test_api_text_to_image.py
```diff
@@ -0,0 +1,67 @@
+import os
```
osanseviero, line 1: Please add a test template to common as well

osanseviero (approved):
Very nice, thanks a lot!

docker_images/diffusers/Dockerfile
Narsil: Docker cache hit. (Subsequent docker builds are much faster)

docker_images/diffusers/app/pipelines/text_to_image.py
Narsil: Why do we need a seed ? By default there's no seed and we just generate at random, and that's good IMO.

docker_images/diffusers/tests/test_api_text_to_image.py
Narsil: It's already there...

PR #79, opened 2022-08-01
ENH sklearn: use columns from config.json

adrinjalali:
This PR makes the API inference to use the columns stored in the config file and pass the data in the right order to the underlying model.

Needs tests, but that requires some pre-work, which is coming in a separate PR to keep this smaller, hence a draft PR for now.

cc @BenjaminBossan


docker_images/sklearn/tests/test_api_tabular_classification.py
```diff
@@ -67,6 +67,18 @@ def test_success_code(self):
...
+        self._check_requirement(self.case_data["old_sklearn"])
+
+        data = self.data
+        with TestClient(self.app) as client:
+            response = client.post("/", json={"inputs": data})
+
+        assert response.status_code == 200
+        content = json.loads(response.content)
+        # TODO: fix this here.
+        assert content == {}
```
adrinjalali, line 80: I've got an issue here. This is the part of the code in `route.py` which attaches warnings to the response:

``` python
    try:
        outputs = pipe(inputs)
        task = os.getenv("TASK")
        metrics = get_metric(inputs, task, pipe)
    except (AssertionError, ValueError) as e:
        outputs = {"error": str(e)}
        status_code = 400
    except Exception as e:
        outputs = {"error": "unknown error"}
        status_code = 500
        logger.error(f"There was an inference error: {e}")

    if warnings and isinstance(outputs, dict):
        outputs["warnings"] = list(sorted(warnings))
```

The issue is they are only attached if there is at least one exception AND some warnings, otherwise warnings are all ignored.

Here's the question:

Should we change `route.py` and if there are only warnings, return a result which includes the output and the warnings, or should my code here raise an error, include the output in the error message for the user to see, and also raise those warnings?

cc @Narsil

adrinjalali: Went the latter route. Not touching `routes.py`.

docker_images/sklearn/app/main.py
```diff
@@ -38,7 +37,6 @@
 }
 
 
-@functools.lru_cache()
```
adrinjalali, line 41: this is removed since `get_pipeline` is always run with no args, and therefore the cache would return the wrong output.

adrinjalali (commented):
This is now ready for a review.

docker_images/sklearn/tests/test_api_tabular_classification.py
```diff
@@ -67,26 +76,90 @@ def test_success_code(self):
...
-        )
+        assert response.status_code == 400
         content = json.loads(response.content)
-        self.assertEqual(set(content.keys()), {"error"})
+        assert "error" in content
+        assert "warnings" in content
+        wrong_versions = [
+            w for w in content["warnings"] if "Trying to unpickle estimator" in w
+        ]
+        assert len(wrong_versions)
```
BenjaminBossan, line 97: Are those lines equivalent to:

`assert any("Trying to unpickle estimator" in w for w in content["warnings"])`

docker_images/sklearn/tests/test_api_tabular_classification.py
```diff
@@ -67,26 +76,90 @@ def test_success_code(self):
...
-            response.status_code,
-            400,
-        )
+        assert response.status_code == 400
+        content = json.loads(response.content)
+        assert "error" in content
+        assert "warnings" in content
+
+        wrong_versions = [w for w in content["warnings"] if warn_message in w]
+        assert len(wrong_versions)
```
BenjaminBossan, line 145: Again, those two lines could be simplified to

`assert any(warn_message in w for w in content["warnings"])`

If not: I think the variable name `wrong_versions` is misleading, since we deal with wrong columns, not wrong versions, right?

docker_images/sklearn/tests/test_api_tabular_classification.py
```diff
@@ -67,26 +76,90 @@ def test_success_code(self):
...
+        assert response.status_code == 400
+        content = json.loads(response.content)
+        assert "error" in content
+        assert "An error occurred while loading the model:" in content["error"]
+
+    @parameterized.expand(
+        [
+            ("add", "The following columns were given but not expected:"),
+            ("drop", "The following columns were expected but not given:"),
+        ]
```
BenjaminBossan, line 119: There could also be a test case for both add and drop.

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,49 +1,129 @@
...
 
 import joblib
+import pandas as pd
 from app.pipelines import Pipeline
 from huggingface_hub import snapshot_download
 
 
 DEFAULT_FILENAME = "sklearn_model.joblib"
 
+logger = logging.getLogger()
```
BenjaminBossan, line 15: Most other modules seem to use `logger = logging.getLogger(__name__)`.

BenjaminBossan (commented):
Thanks for this addition, it's good to have a thorough suite of validation and testing and I think this delivers.

As I'm not very familiar with the api-inference code base as a whole, I cannot really comment on that, so others have to help out there.

I left a few minor suggestions and a request for clarification, so nothing too important. Also, I saw that you are logging the warnings but this is not explicitly tested, not sure if this is expected here or not.

osanseviero (commented):
This looks to be in a good direction, I did an initial light pass

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,49 +1,129 @@
...
-        column_values = list(inputs["data"].values())
-        rows = list(zip(*column_values))
-        # TODO: we should do sth with the warnings.
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore")
-            return self.model.predict(rows).tolist()
+        if getattr(self, "_load_exception", None):
+            # there has been an error while loading the model. We need to raise
+            # that, and can't call predict on the model.
+            raise ValueError(
```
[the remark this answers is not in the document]
adrinjalali: cause we need to have it raised every time the user sends an API request so that they get the right error. This is caught and handled correctly in `routes.py`.

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,49 +1,129 @@
...
+            A :obj:`list` of floats or strings: The classification output for
+            each row.
         """
-        column_values = list(inputs["data"].values())
-        rows = list(zip(*column_values))
-        # TODO: we should do sth with the warnings.
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore")
-            return self.model.predict(rows).tolist()
+        if getattr(self, "_load_exception", None):
```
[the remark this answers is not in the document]
BenjaminBossan: I agree that `hasattr` is better if the default value is not actually being used (except for the conditional).
Alternatively, you could set `self._load_exception = None/False` in `__init__` and then just check `if self._load_exception`.

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,49 +1,129 @@
...
+                "An error occurred while loading the model: "
+                f"{str(self._load_exception)}"
+            )
+
+        _warnings = []
+        if self.columns:
+            # TODO: we should probably warn if columns are not configured, we
+            # really do need them.
+            given_cols = set(inputs["data"].keys())
+            expected = set(self.columns)
```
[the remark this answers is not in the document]
adrinjalali: It's a dict, can't have repeated column names.

docker_images/sklearn/app/pipelines/tabular_classification.py
```diff
@@ -1,49 +1,129 @@
...
+            exception = e
+
+        for warning in record:
+            _warnings.append(f"{warning.category.__name__}({warning.message})")
+
+        for warning in self._load_warnings:
+            _warnings.append(f"{warning.category.__name__}({warning.message})")
+
+        # making sure warnings are recorded only once.
+        _warnings = set(_warnings)
```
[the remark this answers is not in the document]
adrinjalali: I don't think the order should matter, but sure.

docker_images/sklearn/tests/generators/iris.py
```diff
@@ -43,13 +42,18 @@ def push_repo(repo_name, local_repo):
     )
 
 
-if __name__ == "__main__":
-    version = sys.argv[1]
+def get_estimators(X, y):
+    # returns a list of estimators and their names to train and push to hub.
```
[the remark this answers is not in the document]
adrinjalali: Constructing the instances can be more involved than a single liner, that's why I rather have them in a generator.

docker_images/sklearn/tests/generators/iris.py
```diff
@@ -43,13 +42,18 @@ def push_repo(repo_name, local_repo):
...
+    yield "logistic_regression", make_pipeline(
+        StandardScaler(), LogisticRegression()
+    ).fit(X, y)
+    yield "hist_gradient_boosting", HistGradientBoostingClassifier().fit(X, y)
```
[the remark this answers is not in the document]
adrinjalali: added comments, hope it's clear now.

docker_images/sklearn/tests/test_api_tabular_classification.py
adrinjalali: Adding it added complexity to the test, don't love it, but it's there now.

adrinjalali (commented):
Thanks for the reviews. They are address now I think.

docker_images/sklearn/app/pipelines/tabular_classification.py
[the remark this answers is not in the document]
osanseviero: Great, thanks for explaining!

Narsil (approved):
Really like this PR.

The error handling is verbose but seems necessary ! 

Great job.

adrinjalali:
Can this be merged then @BenjaminBossan @osanseviero ?

osanseviero (approved):
Let's ship this! :fire:

BenjaminBossan:
LGTM

adrinjalali:
@Narsil merge and deploy? 😁

Narsil:
Done.

Sorry about the delay !

PR #80, opened 2022-08-02
TST sklearn: better test infrastructure and more tests

adrinjalali:
https://github.com/huggingface/api-inference-community/pull/79 would need a better way to test improvements, but it'd make that PR quite large. This PR creates that infrastructure.

It includes a script which would create the repositories used in tests on the hub as well. It can easily be extended to other tasks and cases.

cc @osanseviero @BenjaminBossan @Narsil


Narsil (approved):
This seems really nice to be able to create easily test repos.

Anything we can do to further help ?

adrinjalali:
> Anything we can do to further help ?

You could merge if there are no red flags :D

osanseviero:
I'll take a look at the PR as well if that's ok :)

adrinjalali:
> I'll take a look at the PR as well if that's ok :)

that'd be really nice!

docker_images/sklearn/tests/generators/iris.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+
+if __name__ == "__main__":
+    version = sys.argv[1]
+
+    X, y = load_iris(return_X_y=True, as_frame=True)
+    X_train, X_test, y_train, y_test = train_test_split(
+        X, y, test_size=0.2, random_state=42
+    )
```
osanseviero, line 56: Why not directly train with X and y? Or if you don't plan to upload metrics, maybe

```suggestion
    X_train, _, y_train, _ = train_test_split(
        X, y, test_size=0.2, random_state=42
    )
```

docker_images/sklearn/tests/generators/iris.py
```diff
@@ -0,0 +1,103 @@
...
+    with open(pkl_name, mode="bw") as f:
+        pickle.dump(est, file=f)
+
+    local_repo = mkdtemp(prefix="skops-")
+    hub_utils.init(
+        model=pkl_name,
+        requirements=[f"scikit-learn={sklearn.__version__}"],
+        dst=local_repo,
+        task="tabular-classification",
+        data=X,
```
osanseviero, line 70: (out of curiosity and totally unrelated to this PR, but what's the use of `data`?)

docker_images/sklearn/tests/test_api_tabular_classification.py
```diff
@@ -38,19 +46,26 @@ def tearDown(self):
...
-        self.assertEqual(
-            response.status_code,
-            200,
-        )
+
+        assert response.status_code == 200
         content = json.loads(response.content)
-        self.assertEqual(type(content), list)
-        self.assertEqual(len(content), expected_output_len)
+        assert type(content) == list
```
osanseviero, line 67: Why are you using asserts instead of `self.assertEqual` for consistency with the rest of this repo's codebase?

osanseviero (approved):
Very nice and useful! Huge benefit of being able to train models in seconds with `sklearn` :D I left some minor comments but this looks great

docker_images/sklearn/tests/generators/iris.py
adrinjalali: this is used partly to shuffle the data. And the `X_test` is used bellow.

docker_images/sklearn/tests/generators/iris.py
adrinjalali: two things:
- get the first three rows to save in `config.json` and the README metadata's `widget` section for widgets to work.
- stores the columns and their order in `config.json` so that the bakend can construct a correct data to be passed to the model. See https://github.com/huggingface/api-inference-community/pull/79/files

docker_images/sklearn/tests/test_api_tabular_classification.py
adrinjalali: because we use `pytest` and it correctly understands `assert`. I find `assert x == y` significantly more readable and elegant than `self.assertEqual(x, y)`.

"We should do it the XXX way cause we've done it this way" is not really a good argument ;)

docker_images/sklearn/tests/generators/iris.py
osanseviero: Ah alright! It was passing `X` before hehe

docker_images/sklearn/tests/generators/iris.py
```diff
@@ -0,0 +1,103 @@
...
+from sklearn.linear_model import LogisticRegression
+from sklearn.model_selection import train_test_split
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+from skops import hub_utils
+
+
+def push_repo(repo_name, local_repo):
+    # this token should be allowed to push to the skops-tests org.
+    token = os.environ["SKOPS_TESTS_TOKEN"]
```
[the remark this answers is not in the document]
adrinjalali: The readme says: 


> You can make sure those repos are up to date by running the
> `docker_images/sklearn/tests/generators/run.sh` script. The script creates
> required conda environments and runs scripts inside those environments. You
> should also give it a valid token with access to the `skops-tests` org:
> 
> ```
> SKOPS_TESTS_TOKEN=your_secret_token path/to/run.sh
> ```

The user wouldn't run the python file, the script runs it.

docker_images/sklearn/tests/test_api_tabular_classification.py
osanseviero: Consistency in a codebase is more important to me, especially as other library owners look into other implementations when implementing theirs. Up to you though!

adrinjalali: They should start using `assert` too :D

Narsil: Actually no, `assert` is not great because we cannot enforce the use `pytest`. `python -m unittest` works just as well.
And if you're not using it, then the error messages are pretty unreadable.

Also `transformers` is also using that format, and pretty much every lib in the HF ecosystem I have seen so far.
It's just the most portable one, so let's go for this one, please.

I won't strongly oppose this as long as the test does its job, but I feel more favorable to `assertEqual`.

docker_images/sklearn/tests/generators/iris.py
osanseviero: Very cool, thanks for explaining!

docker_images/sklearn/tests/test_api_tabular_classification.py
adrinjalali: Note that these tests do use pytest. It's importing it, so w/o `pytest` this PR doesn't work anyway. And we pretty much never ask people to run tests with unittest, we always give them a pytest command to run.

Narsil: If I am not mistaken, you are adding the hard dependency, otherwise the file would be relatively agnostic.

You are free to break consistency here as `sklearn` inference is meant to be isolated from the rest, just commenting about the rest of HF ecosystem.

I'll resolve as I think there's not really anything to add anymore.

adrinjalali:
Shall we merge this?

osanseviero:
Yes, merging!

PR #83, opened 2022-08-17
Add tabular regression

BenjaminBossan:
Add tabular regression as a new task type. Fixes https://github.com/skops-dev/skops/issues/100

## Description

This PR mostly mirrors tabular classification when it comes to implementation (same pipeline, almost same testing). There is some code duplication because of this, LMK if you prefer to have that abstracted away.

## Missing

1. Generic tests

~There are some places where I believe generic tests have to be added but I was not certain (see `TODO` comments). Please LMK if I should add them there.~ After discussion with Adrin, this PR is not the right place to add them.

2. Documentation

I didn't find a place in this repo that documents the existing tasks. Please LMK if I overlooked it.


BenjaminBossan:
@adrinjalali Could you give this a first pass to check if there is nothing major missing?

docker_images/generic/tests/test_api.py
```diff
@@ -11,6 +11,7 @@
     "feature-extraction": ["osanseviero/fasttext_english"],
     "image-classification": ["osanseviero/fastai_cat_vs_dog"],
     "tabular-classification": ["osanseviero/wine-quality"],
+    # TODO "tabular-regression": [],
```
adrinjalali, line 14: why are you touching this under `generic`?

docker_images/sklearn/tests/generators/iris.py
```diff
@@ -26,9 +28,15 @@ def push_repo(repo_name, local_repo):
     print(f"Pushing {repo_id}")
 
     client = HfApi()
-    client.delete_repo(repo_id, token=token)
+    try:
+        client.delete_repo(repo_id, token=token)
+    except RepositoryNotFoundError:
```
adrinjalali, line 33: If we can't import this exception from a public path from `huggingface_hub`, I rather catch `Exception` instead.

Or to check if the repo exists explicitly

docker_images/sklearn/tests/generators/iris.py
```diff
@@ -26,9 +28,15 @@ def push_repo(repo_name, local_repo):
...
+        # repo does not exist yet
+        pass
     client.create_repo(repo_id=repo_id, token=token, repo_type="model")
 
+    # prevent AWS "503 Server Error: Slow Down for url" error
+    time.sleep(10)
```
adrinjalali, line 39: I've never had an issue here though. 10 seconds is a lot of time, maybe add like half a second if you really need to? This script never failed for me.

docker_images/sklearn/tests/generators/tabularregression.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
+
+# This is an amost 1:1 copy of iris.py but uses regression instead of
```
adrinjalali, line 3: should we then just put all of this in one single file?

setup.py
```diff
@@ -17,7 +17,6 @@
         "test": [
             "httpx>=0.18",
             "Pillow>=8.2",
-            "httpx>=0.18",
```
adrinjalali, line 20: why are you touching this file?

tests/test_nlp.py
```diff
@@ -161,11 +161,10 @@ def test_valid_input(self):
 
         inputs = {"data": data}
         bpayload = json.dumps({"inputs": inputs}).encode("utf-8")
-        normalized_inputs, processed_params = normalize_payload_nlp(
-            bpayload, "tabular-classification"
-        )
-        self.assertEqual(processed_params, {})
-        self.assertEqual(inputs, normalized_inputs)
+        for task in ["tabular-classification", "tabular-regression"]:
```
adrinjalali, line 164: also, why touching this file?

adrinjalali (commented):
This looks quite good to me. I'm impartial on whether to deduplicate the test code or not, but I think it'd be simpler to merge this if it didn't touch anything outside the sklearn folder.

docker_images/sklearn/tests/generators/iris.py
BenjaminBossan: I'm changing this to `from huggingface_hub.utils import RepositoryNotFoundError`.

docker_images/sklearn/tests/generators/iris.py
BenjaminBossan: I tried lower times and at 7 seconds it still failed, so I had to go to 10. I have no idea why this would cause the error, as we are just performing a few operations, but maybe they resolve to many S3 operations under the hood?

I agree that 10 sec is very long but since this script only needs to be called quite rarely, I thought we can live with that (exponential backoff would be more efficient).

docker_images/sklearn/tests/generators/tabularregression.py
BenjaminBossan: Do you mean to put it into a single script as is, or try to make it more generic so that some functions can be re-used?

setup.py
BenjaminBossan: Boyscout rule ;) I can revert if it's not desired.

tests/test_nlp.py
BenjaminBossan: I got the impression that since tabular classification is tested here and now we add tabular regression, it should be tested in the same way.

docker_images/sklearn/tests/generators/iris.py
adrinjalali: Sure, then we leave it as is for now.

cc @julien-c, I'm not sure if this is expected behavior.

docker_images/sklearn/tests/generators/tabularregression.py
adrinjalali: I think with a bit of tweak the code in these two files can be made a bit more generic to avoid duplication.

setup.py
adrinjalali: I'd move it to a separate PR then.

tests/test_nlp.py
adrinjalali: I agree, but I'd put it in a separate PR.

docker_images/generic/tests/test_api.py
BenjaminBossan: Reverted.

docker_images/sklearn/tests/generators/tabularregression.py
BenjaminBossan: Done, please take a look at `generate.py`

setup.py
BenjaminBossan: Reverted.

tests/test_nlp.py
BenjaminBossan: Reverted.

docker_images/sklearn/tests/generators/generate.py
```diff
@@ -0,0 +1,229 @@
...
+    # prevent AWS "503 Server Error: Slow Down for url" error
+    time.sleep(SLEEP_BETWEEN_PUSHES)
+
+
+def get_tabularregression_data():
+    # Create an artificial regression dataset similar in structure to Iris data
+    n_features = 4
+    X, y = make_regression(150, n_features=n_features, n_informative=2, random_state=42)
+    columns = [f"col{i}" for i in range(n_features)]
+    return pd.DataFrame(X, columns=columns), pd.Series(y)
```
adrinjalali, line 77: Let's use `load_diabetes` instead, more meaningful features and just a single liner to get the data.

docker_images/sklearn/tests/generators/generate.py
```diff
@@ -0,0 +1,229 @@
...
+
+def predict_tabular_regressor(est, sample, filename):
+    output = [float(x) for x in est.predict(sample)]
+    with open(Path(__file__).parent / "samples" / filename, "w") as f:
+        json.dump(output, f, indent=2)
+
+
+#############
+# CONSTANTS #
+#############
```
adrinjalali, line 166: move to top of the file?

adrinjalali (commented):
minor nits, otherwise LGTM!

docker_images/sklearn/tests/generators/generate.py
BenjaminBossan: Doesn't work since the constants partly reference functions that need to be defined before.

BenjaminBossan (commented):
I changed the dataset as requested.

adrinjalali (approved):

adrinjalali:
@Narsil @osanseviero would you like to have a look?

docker_images/sklearn/app/pipelines/tabular_regression.py
```diff
@@ -0,0 +1,6 @@
+from app.pipelines.tabular_classification import TabularClassificationPipeline
```
osanseviero, line 1: Could you also add a template in https://github.com/huggingface/api-inference-community/tree/main/docker_images/common/app/pipelines please? This can be the same as in https://github.com/huggingface/api-inference-community/blob/main/docker_images/common/app/pipelines/tabular_data_pipeline.py

osanseviero (approved):
This looks great, thanks for adding this!

So this is useful for other libraries, could you also update https://github.com/huggingface/api-inference-community/tree/main/docker_images/common? `common` is copy-pasted when starting to implement pipelines for other libraries.

Apart from it, I would also suggest to add a docker test at https://github.com/huggingface/api-inference-community/blob/main/tests/test_dockers.py#L149

BenjaminBossan:
Thanks for the feedback @osanseviero 

> So this is useful for other libraries

Could you please specify what *this* refers to? Not sure what part you mean.

> Could you also add a template

I assume the intent is to have one class per file here? So that means that `tabular_data_pipeline.py` should be split into `tabular_classification_pipeline.py` and `tabular_regression_pipeline.py` (and classes renamed accordingly)?

adrinjalali:
should we implement the suggestions about other subfolders in a separate PR?

osanseviero:
> should we implement the suggestions about other subfolders in a separate PR?

As you prefer :) except the docker test, which should go in hand with the skops change https://github.com/huggingface/api-inference-community/blob/main/tests/test_dockers.py#L149

> I assume the intent is to have one class per file here? So that means that tabular_data_pipeline.py should be split into tabular_classification_pipeline.py and tabular_regression_pipeline.py (and classes renamed accordingly)?

Yes, splitting them (in /common) would make sense as we split the task into two.

> Could you please specify what this refers to? Not sure what part you mean.

When adding new libraries to the API, developers usually copy-paste the code from /common. When we add new pipeline types as you're doing in this PR (adding `tabular-regression`), we usually also update /common with a template pipeline (you can check out how it was done for [TabularDataPipeline](https://github.com/huggingface/api-inference-community/blob/main/docker_images/common/app/pipelines/tabular_data_pipeline.py) and the tests which should look similar to the ones you're adding in this PR. (I see there is a test [here](https://github.com/huggingface/api-inference-community/blob/main/docker_images/common/tests/test_api_tabular_regression.py) already, so it might be unneeded). As mentioned by @adrinjalali above this could be a separate PR, but ideally should be done in parallel, so we don't have an outdated `/common`.

BenjaminBossan:
I addressed the comments

- Rename `TabularDataPipeline` to `TabularClassificationPipeline` (including
  module name)
- Adjust some types to include int
- Add `TabularRegressionPipeline`
- Add docker test for tabular regression

> When adding new libraries to the API, developers usually copy-paste the code from /common. When we add new pipeline types as you're doing in this PR (adding tabular-regression), we usually also update /common with a template pipeline

Thank you for explaining.

> see there is a test [here](https://github.com/huggingface/api-inference-community/blob/main/docker_images/common/tests/test_api_tabular_regression.py) already, so it might be unneeded

Yes, it looks like there was nothing left to do there.

tests/test_dockers.py
```diff
@@ -146,13 +146,21 @@ def test_flair(self):
...
 
-    def test_sklearn(self):
+    def test_sklearn_tabular_classification(self):
         self.framework_docker_test(
             "sklearn",
             "tabular-classification",
             "julien-c/wine-quality",
         )
 
+    def test_sklearn_tabular_regression(self):
```
Narsil, line 156: Can we keep the test in a single `test_sklearn` ?

These are very high level integration test, it makes no sense to split them since every single model should pass, or something is wrong (I mostly use them to sanity check before deploying)

BenjaminBossan: Done

Narsil:
May I merge&deploy ?

BenjaminBossan:
From my point of view, there is nothing more to add. Not sure if the other reviewers want to take another look or not.

Narsil:
I'll merge, the other reviewers approved. I fanything is missing we can add later.

Narsil:
Yup, I was waiting for the fix to the docker themselves https://github.com/huggingface/api-inference-community/pull/85

Since the docker test was not passing I wasn't sure we should update yet.

Narsil:
Proded

PR #85, opened 2022-08-22
Sklearn docker tests update

BenjaminBossan:
sklearn tests (tabular classification and regression) failed because the underlying model was using an older version (0.24). Now we use the latest version created via the `docker_images/sklearn/tests/generatos/run.sh` script.

Furthermore, for inference calls to these models to work, we need specific input data, which can now optionally be passed to the `framework_docker_test` function.


adrinjalali (approved):
Nice fix!

Narsil (approved):
LGTM.

PR #89, opened 2022-08-29
Fixing pyannote.audio integration

hbredin:
pyannote.audio integration seems to be broken: models takes forever to load, ending up in a time out.

This is a (desperate) attempt at fixing this...

image

### DIFF
--- a/docker_images/pyannote/requirements.txt
+++ b/docker_images/pyannote/requirements.txt
@@ -1,5 +1,9 @@
 starlette==0.14.2
-api-inference-community==0.0.23
-pyannote-audio==2.0.1
+api-inference-community==0.0.25
+torch==1.11.0
+torchvision==0.12.0
+torchaudio==0.11.0
+torchtext==0.12.0
 speechbrain==0.5.12
-huggingface_hub==0.5.1
+pyannote-audio==2.0.1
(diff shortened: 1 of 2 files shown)


hbredin:
Looks like one Docker test is failing (that used to not fail)...

DockerImageTests.test_pyannote ________________________

self = <test_dockers.DockerImageTests testMethod=test_pyannote>

```
    def test_pyannote(self):
>       self.framework_docker_test(
            "pyannote",
            "automatic-speech-recognition",
            "pyannote/voice-activity-detection",
        )
E       TypeError: framework_docker_test() missing 1 required positional argument: 'custom_input'
```

I have no idea what this `custom_input` should be...

osanseviero:
Sorry, that was an unrelated bug. I just did a commit to see if it fixes the issue

Narsil:
Oh the issue is that the docker image is not found because it's misconfigured. 

I think this is because the library_name of the hub is `pyannote_audio` and not `pyannote`  (here in this library it's defined as `pyannote`.)

Narsil:
It's working now ! https://huggingface.co/pyannote/speaker-diarization

hbredin:
Thanks a lot for the quick fix!

PR #91, opened 2022-08-29
update the fairseq version to support more models

lpw0:
Updating the [version](https://github.com/facebookresearch/fairseq/commit/eda703798dcfde11c1ee517805c27e8698285d71) of fairseq to test new models in the [list](https://github.com/facebookresearch/fairseq/blob/d81fac8163364561fd6cd9d82b6ee1ba502c3526/fairseq/models/speech_to_text/xm_transformer.py#L493).


Narsil (approved):
LGTM.

The timeout in test always existed (seems there's a lot to load even when we hit disk cache)

Narsil:
In prod.

PR #100, opened 2022-09-14
Add support for SpeechBrain SoundChoice G2P (core part)

flexthink:
Hosted Interface API for the SoundChoice G2P model (core part)

https://huggingface.co/speechbrain/soundchoice-g2p
https://arxiv.org/abs/2207.13703

### DIFF
--- a/api_inference_community/validation.py
+++ b/api_inference_community/validation.py
@@ -189,7 +189,7 @@ INPUTS_MAPPING = {
     "zero-shot-classification": StringInput,
     "text-to-speech": StringInput,
     "text-to-image": StringInput,
-    "grapheme-to-phoneme": StringInput
+    "text-to-text": StringInput
 }
 
 BATCH_ENABLED_PIPELINES = ["feature-extraction"]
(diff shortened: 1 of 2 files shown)


flexthink:
This needs to be merged first (otherwise the grapheme-to-phoneme task will not be allowed by validation):
https://github.com/huggingface/api-inference-community/pull/100

osanseviero:
Hey there! For simplicity and given this is the first model on the Hub for this task, instead of creating a new type `grapheme-to-phoneme`, maybe we can reuse `text2text-generation`. This has some benefits:

* It has same input and output types (text to text)
* The widget and API would work out of the box

This would be easier to maintain and we could launch it very quickly. One last comment, instead of joining with a space, maybe we can use a hyphen (-) so the output string is like "T-UW- -B-IY" as some phonemes can be empty strings

```python
from speechbrain.pretrained import GraphemeToPhoneme
g2p = GraphemeToPhoneme.from_hparams("speechbrain/soundchoice-g2p")
text = "To be or not to be, that is the question"
phonemes = g2p(text)
# ['T', 'UW', ' ', 'B', 'IY', ' ', 'AO', 'R', ' ', 'N', 'AA', 'T', ' ', 'T', 'UW', ' ', 'B', 'IY', ' ', 'DH', 'AE', 'T', ' ', 'IH', 'Z', ' ', 'DH', 'AH', ' ', 'K', 'W', 'EH', 'S', 'CH', 'AH', 'N']
```

flexthink:
This is no longer needed - after reusing text2text-generation

PR #107, opened 2022-09-20
First step towards CD. (Only flair for now)

Narsil:
The goal is to automatically deploy on merge.
The helm prod chart lives in another repo, ideally we trigger
another action in this other repo (it's simpler). 

The building and uploading can happen here, we just need to add some secrets to this repo.

Feedback welcome on the approach.

Ideally I would like to have a single CD action, that would just check at the modified folders and deploy every folder hit in `docker_images`.

This seems doable however not highly stable (depends on the trigger of the action to get the correct git commands)

Requires: https://github.com/huggingface/api-inference/pull/1151

### DIFF
--- a/.github/workflows/python-api-flair-cd.yaml
+++ b/.github/workflows/python-api-flair-cd.yaml
@@ -1,10 +1,10 @@
 name: flair-docker-cd
 on:
   push:
-    # branches:
-    #   - main
-    # paths:
-    #   - "docker_images/flair/**"
+    branches:
+      - main
+    paths:
+      - "docker_images/flair/**"
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/python-api-flair-cd.yaml
+++ b/.github/workflows/python-api-flair-cd.yaml
@@ -22,14 +22,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install --upgrade pip
-          pip install kubernetes httpx uvicorn awscli
-          sudo snap install kubectl --classic
-          helm plugin install https://github.com/databus23/helm-diff
-      - name: Setup K8S
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/python-api-flair-cd.yaml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Install dependencies
```
OlivierDehaene, line 22: I don't think we need kubectl, helm kubernetes and uvicorn(?)

.github/workflows/python-api-flair-cd.yaml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          pip install --upgrade pip
+          pip install kubernetes httpx uvicorn awscli
+          sudo snap install kubectl --classic
+          helm plugin install https://github.com/databus23/helm-diff
+      - name: Setup K8S
```
OlivierDehaene, line 28: Should not be needed.

osanseviero (commented):
Overall looks like a good approach, thanks a lot for working on this!

.github/workflows/python-api-flair-cd.yaml
Narsil: Thanks for the catch !

.github/workflows/python-api-flair-cd.yaml
Narsil: Thanks for the catch !

OlivierDehaene (approved):

PR #108, opened 2022-09-20
Fix SpeechBrain text generation

osanseviero:
This also adds templates to common to avoid this in the future


Narsil (approved):
LGTM.

I hope this is consistent with the widget. but `transformers` is not necessarily hundred consistent with itself (like `text-generation` will return a list, or not, depending on parameters, or worse sometimes, the payload itself).

Thanks !

PR #112, opened 2022-09-27
Adding fairseq CD.

Narsil:


osanseviero (approved):
Approving as it's same as https://github.com/huggingface/api-inference-community/pull/107 but for Fairseq

PR #114, opened 2022-09-28
Add warnings to error response for sklearn Pipelines

BenjaminBossan:
See #92

When calling the prediction pipeline produces warnings, it is treated as an error. This includes situations where the model output *might* be correct but we're not sure, most notably when there is a mismatch of the version of the loaded sklearn model and the sklearn version of the runtime.

Although we still return the model outputs to the user, we don't show them what warnings were actually encountered. This makes debugging difficult. Therefore, we want to show them the warnings.

This requires two changes. First, we need to add the encountered warnings to the response. Second, we need to display these warnings. This PR implements the first step.

On top of that, this PR introduces a small refactoring to reduce the amount of code duplication (see the last paragraph of [this comment](https://github.com/huggingface/api-inference-community/pull/92#issuecomment-1231776124)). This refactoring results in no functional changes.

### DIFF
--- a/docker_images/sklearn/app/pipelines/common.py
+++ b/docker_images/sklearn/app/pipelines/common.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import warnings
+from abc import abstractmethod
 from pathlib import Path
 from typing import Any
 
@@ -60,17 +61,21 @@ class SklearnBasePipeline(Pipeline):
         # to the model in the right order.
         self.columns = config.get("sklearn", {}).get("columns", None)
 
+    @abstractmethod
     def _get_output(self, inputs: Any) -> Any:
         raise NotImplementedError(
             "Implement this method to get the model output (prediction)"
         )
 
-    def handle_call(self, inputs: Any) -> Any:
+    def __call__(self, inputs: Any) -> Any:
         """Handle call for getting the model prediction
 
         This method is responsible for handling all possible errors and
         warnings. To get the actual prediction, implement the `_get_output`
         method.
+
+        The types of the inputs and output depend on the specific task being
+        implemented.
 
         """
         if self._load_exception:
--- a/docker_images/sklearn/app/pipelines/common.py
+++ b/docker_images/sklearn/app/pipelines/common.py
@@ -5,9 +5,8 @@ from pathlib import Path
 from typing import Any
 
 import joblib
-from huggingface_hub import snapshot_download
-
 from app.pipelines import Pipeline
+from huggingface_hub import snapshot_download
 
 
 logger = logging.getLogger(__name__)
@@ -24,6 +23,7 @@ class SklearnBasePipeline(Pipeline):
       annotations.
 
     """
+
     def __init__(self, model_id: str):
         cached_folder = snapshot_download(repo_id=model_id)
         self._load_warnings = []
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
Not sure why there is a timeout error. Log says:

> timed out after 20 seconds

even though the call seems to be:

```python
        self.framework_docker_test(
            "sklearn",
            "tabular-classification",
            "skops-tests/iris-sklearn-latest-logistic_regression-with-config",
            custom_input=clf_data,
            timeout=600,
        )
```

The whole CI fails after 3 min, so these 600 seconds don't seem to be honored.

docker_images/sklearn/app/pipelines/common.py
```diff
@@ -46,3 +59,78 @@ def __init__(self, model_id: str):
         # use column names from the config file if available, to give the data
         # to the model in the right order.
         self.columns = config.get("sklearn", {}).get("columns", None)
+
+    def _get_output(self, inputs: Any) -> Any:
+        raise NotImplementedError(
+            "Implement this method to get the model output (prediction)"
+        )
+
+    def handle_call(self, inputs: Any) -> Any:
```
adrinjalali, line 68: to me it seems that this is actually `__call__`. So why not implement it here, and remove the one line implementation form the child classes, and only have them implement `_get_output`?

adrinjalali (commented):
What's the change here which fixes the original issue of not returning the warnings?

docker_images/sklearn/app/pipelines/common.py
BenjaminBossan: > So why not implement it here, and remove the one line implementation form the child classes

I tried to explain it in the comment:

```python
    # even though we only delegate the call, we implement this method have the
    # correct docstring and type annotations
```

(just noticed a missing word there)

That's also why I renamed the method in the parent class -- this way, the concrete classes cannot forget to implement `__call__`. If we don't care for the annotations and docstring, I can do the change you suggested.

BenjaminBossan:
> What's the change here which fixes the original issue of not returning the warnings?

https://github.com/huggingface/api-inference-community/pull/114/files#diff-e20608fb08c5272baa6ed6abca2e2a940ee3b6ff51f1088b9d04970cb681f083R125

docker_images/sklearn/app/pipelines/common.py
adrinjalali: But we don't really care about the typing and docstring here, do we? these classes are never used by end users or other devs.

BenjaminBossan: Maybe, I don't know if and how they're used. I just saw that they were there so I assumed they served a purpose. If they don't, I'm happy to remove this indirection. Maybe @Narsil can comment on that.

Narsil: Oh no the scaffolding aims to kickstart new implementation, then you're free to do whatever you want.

Also the parent class should already yell if `__init__` or `__call__` or not implemented.

Narsil:
>     timed out after 20 seconds

This is when the docker wants to be shut off. I tried manually to see what was happening but it's true that the docker image doesn't see to honor the SIGTERM signal. I don't really understand what's going on since all other docker seem to be working just fine.

The only thing that seems to work locally is to forcefully discard the container.
On the cluster it's OK since k8s already does this, but it's still very odd/peculiar behavior.

BenjaminBossan:
@adrinjalali @Narsil I addressed your comments.

Timeout unfortunately still occurs.

PR #115, opened 2022-09-28
fix generation issue and remove unit text for s2ut models

lpw0:
Some changes in this commit:
- Add `beam` and `max_len_a` configuration to fix generation issue to avoid duration limitations of translation results.
- Replace the unit text output of S2UT models with an empty string. 
- Remove `cpu=true`, local test passed with `pytest -sv --rootdir docker_images/fairseq/ docker_images/fairseq/` , maybe need to further check after it is deployed.


Narsil (approved):
LGTM

osanseviero (approved):

PR #116, opened 2022-10-05
Update flair version to support latest models

irg1008:
Update flair version to support word and transformer embeddings on widget and inference API.

Some models are only working with latest versions of flair.

For example: This [example of flair 0.11.3 model](https://huggingface.co/lighthousefeed/yoda-ner) only works with latest commit of the library.
You can check it here: [colab notebook](https://colab.research.google.com/drive/1EnL0IFLVN1FDO1x42r13A8oZsfdBcHPg?usp=sharing)

### DIFF
--- a/docker_images/flair/requirements.txt
+++ b/docker_images/flair/requirements.txt
@@ -1,5 +1,4 @@
 starlette==0.14.2
 pydantic==1.8.1
 flair @ git+https://github.com/flairNLP/flair@d4ed67bf663e4066517f00397412510d90043653
-api-inference-community==0.0.23
-huggingface_hub==0.5.1
+api-inference-community==0.0.25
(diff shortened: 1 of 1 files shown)


docker_images/flair/requirements.txt
```diff
@@ -1,5 +1,5 @@
 starlette==0.14.2
 pydantic==1.8.1
-flair==0.11.1
+flair @ git+https://github.com/flairNLP/flair@d4ed67bf663e4066517f00397412510d90043653
 api-inference-community==0.0.23
 huggingface_hub==0.5.1
```
osanseviero, line 5: There seems to be an installation error as different `huggingface_hub` versions are expected for flair, so this should fix it

```suggestion
api-inference-community==0.0.25
```

osanseviero (approved):
Thanks a lot! I had a minor issue installing the requirements, but otherwise the tests pass!

PR #121, opened 2022-10-13
Support speech_to_speech task in audio_to_audio pipeline

cndn:
This PR contains the change to enable facebook/textless_sm_cs_en. It's with a separate fairseq task and hub interface. Tests passed on my laptop when I switch TESTABLE_MODELS from facebook/xm_transformer_sm_all-en to facebook/textless_sm_cs_en (Shall we add both for testing going forward?)


Narsil (approved):
Is there any reason you can't update the model directly on the hub ?

And implement speech_to_text directly for `"facebook/xm_transformer_s2ut_800m-es-en-st-asr-bt_h1_2022"` ?

I'll approve so we can fix faster in the meantime.

@osanseviero FYI

Narsil:
Do you mind fixing style with `make style`  ?

cndn:
> Is there any reason you can't update the model directly on the hub ?
> 
> And implement speech_to_text directly for `"facebook/xm_transformer_s2ut_800m-es-en-st-asr-bt_h1_2022"` ?
> 
> I'll approve so we can fix faster in the meantime.
> 
> @osanseviero FYI

(1) Good call! Later we could merge S2SHubInterface / S2THubInterface and move the heavy lifting to fairseq side and keep audio-to-audio pipeline generic. 
(2) @lpw0 Do you think we could just update cfg in checkpoint? I remember this is just an one-off case
(3) Will Fix make style

Thanks @Narsil! BTW our model loading never finishes, would you mind helping take a look? Described the issue in our slack channel.

cndn:
Will fix one thing around overrides in 30 mins.

osanseviero (approved):
Thanks for this! As discussed before, as a follow-up it would be great to fix this directly in the repos side to give a consistent experience to users loading the models directly

PR #124, opened 2022-10-14
Triggering a nemo CD.

### LINKED ISSUE #122: Update huggingface nemo docker version
**Is your feature request related to a problem? Please describe.**
Squeezeformer arcitecture support was added in `nemo_toolkit==1.12.0`
But looks like huggingface docker has `nemo_toolkit==1.10.0`

Without it, I can't tun my model via inference API
https://huggingface.co/theodotus/stt_uk_squeezeformer_ctc_ml

**Describe the solution you'd like**
Here in nemo docker images there is no precise version of nemo to install
So it's server side docker conteiner update required 
https://github.com/huggingface/api-inference-community/tree/main/docker_images/nemo

Narsil:
NEed to be merged after #123 

Fixes #122

### DIFF
--- a/docker_images/nemo/requirements.txt
+++ b/docker_images/nemo/requirements.txt
@@ -1,4 +1,5 @@
 starlette==0.14.2
 api-inference-community==0.0.23
+#Dummy update to retrieve latest version
 nemo_toolkit[all]
 huggingface_hub==0.5.1


osanseviero (approved):
Thanks! Maybe let's also allow running the workflows manually to avoid having to do this

Narsil:
Hmmm, if we don't change the file, the docker cache could hit. 

Docker sees everything is the same file, it will just reuse its cache instead of retriggering a download of the `requirements.txt`.
We could add the `workflow_dispatch` for manual clicking, but I feel keeping it into PRs and consistent workflow make it easier to run.
It will also rerun the tests which could be super important to make sure new releases are not broken. (well `nemo` is lacking some tests but hey..)

osanseviero:
Sounds good, thanks for explaining!

PR #132, opened 2022-10-19
Dummy reversion (CD still had a typo).

Narsil:


Narsil:
> Would an empty commit also work to trigger CD?

No the CD triggers when files in the `docker/sklearn` directory are modified.
@osanseviero suggested we could manually trigger the CD, but I think it's slightly better to do it this way because of docker caching which might be hit and prevent a correct deployment from happening (it also just looks at the file, and if it's the same file it just reuses what it had).

Also creating PRs forces the CI to run, so hopefully we can catch more bugs that way.

BenjaminBossan:
Okay, thanks for explaining.

BenjaminBossan (approved):

PR #144, opened 2022-11-17
Fixing build_docker bug.

Narsil:


docker_images/diffusers/requirements.txt
```diff
@@ -8,3 +8,4 @@ pydantic==1.8.1
 ftfy==6.1.1
 # triton==1.1.1
 # -e git+https://github.com/facebookresearch/xformers@7e4c02c#egg=xformers
+# Dumb change.
```
Narsil, line 11: We trigger auto deploys when the source code is modified. This is a bit silly for this particular use case when the earlier build failed but it works.

PR #146, opened 2022-11-17
Fixing diffusers (we need transformers main to fix the CLIPProcessor issue).

Narsil:
This is already live, but cleaning things up and this PR will check that the CD finally works properly.


anton-l (approved):
Makes sense, sorry for the confusion :sweat_smile:

PR #147, opened 2022-11-18
Integrating PaddleNLP to HF Inference API

sijunhe:
- Docker build successfully. Manually tested the generated paddlenlp docker image 
- Todo:  add `fill-mask` task once the paddlenlp can directly load from hf hub

### DIFF
--- a/docker_images/paddlenlp/app/pipelines/conversational.py
+++ b/docker_images/paddlenlp/app/pipelines/conversational.py
@@ -7,14 +7,12 @@ from paddlenlp.taskflow import Taskflow
 
 class ConversationalPipeline(Pipeline):
     def __init__(self, model_id: str):
-        # TODO: how to setup params
         self.pipeline = Taskflow("dialogue", task_path=model_id, from_hf_hub=True)
-        
 
     def __call__(self, inputs: Dict[str, Union[str, List[str]]]) -> Dict[str, Any]:
         """
         Args:
-            inputs (:obj:`dict`): a dictionnary containing the following key values:
+            inputs (:obj:`dict`): a dictionary containing the following key values:
                 text (`str`, *optional*):
                     The initial user input to start the conversation. If not provided, a user input needs to be provided
                     manually using the [`~Conversation.add_user_input`] method before the conversation can begin.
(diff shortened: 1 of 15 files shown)


Narsil:
Thank you for this ! It would definitely be a great addition.

Tell us if you need any help ! .

sijunhe:
Hi @Narsil. Thanks for offering help. Turns out I really do need it. :)
1. See #153 
2. In the repo I see classes for `Params` that holds stuff like `top_k` or `max_length` but I don't see how that get passed into the pipeline or get used. Could you share an example?

Narsil:
> In the repo I see classes for Params that holds stuff like top_k or max_length but I don't see how that get passed into the pipeline or get used. Could you share an example?

Indeed they are not.
The reason is that because of so many different frameworks, we started conservative for parameters and just ignore them because their management is likely to be framework dependant, and even in `transformers`, parameters were not necessarily aligned, and could change the actual pipeline output.

The code is simply leftover from porting the `transformers` inference code (or what it was at the time).

Adding parameters, is definitely something we can revisit but it feels like a larger decision needs to be taken, like:
- Can all framework implement whatever parameter they feel like
- Can parameters modify the output type ?
- Can we have a minimal (optional) subset of parameters that most framework can agree upon ?

The answer to these will put the direction on very different tracks and I think any direction have pros/cons. I think it's a discussion better left for somewhere else.

> See Provide a complete list of tasks in docker_images/common/app/pipelines #153

I'll answer over there.

Narsil:
For the quality test `make style` should fix it.

docker_images/paddlenlp/app/pipelines/conversational.py
```diff
@@ -0,0 +1,55 @@
...
+                    manually using the [`~Conversation.add_user_input`] method before the conversation can begin.
+                past_user_inputs (`List[str]`, *optional*):
+                    Eventual past history of the conversation of the user. You don't need to pass it manually if you use the
+                    pipeline interactively but if you want to recreate history you need to set both `past_user_inputs` and
+                    `generated_responses` with equal length lists of strings
+                generated_responses (`List[str]`, *optional*):
+                    Eventual past history of the conversation of the model. You don't need to pass it manually if you use the
+                    pipeline interactively but if you want to recreate history you need to set both `past_user_inputs` and
+                    `generated_responses` with equal length lists of strings
+        Return:
```
osanseviero, line 27: Nice!

osanseviero (commented):
Thanks a lot! This is going in the right direction 🔥 

Some things
- The fill-mask task does not seem supported in the library yet
- Could we add GitHub Actions workflows in https://github.com/huggingface/api-inference-community/tree/main/.github/workflows ?  One for CD as in https://github.com/huggingface/api-inference-community/blob/main/.github/workflows/python-api-nemo-cd.yaml and another for testing the image as in https://github.com/huggingface/api-inference-community/blob/main/.github/workflows/python-api-sklearn.yaml
- It seems like you have some quality errors. Please run `make quality` to make sure this is properly formatted

Thanks again 🚀

docker_images/paddlenlp/app/pipelines/fill_mask.py
```diff
@@ -0,0 +1,23 @@
+from http.client import responses
+from typing import Any, Dict, List, Union
+
+from app.pipelines import Pipeline
+from paddlenlp.taskflow import Taskflow
+
+
+class FillMaskPipeline(Pipeline):
+    def __init__(self, model_id: str):
+        self.taskflow = Taskflow("fill_mask", task_path=model_id, from_hf_hub=True)
```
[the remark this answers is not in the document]
sijunhe: I just merged it. https://github.com/PaddlePaddle/PaddleNLP/pull/3870

docker_images/paddlenlp/requirements.txt
```diff
@@ -0,0 +1,6 @@
+starlette==0.14.2
+api-inference-community==0.0.23
+huggingface_hub==0.5.1
+paddlepaddle==2.3.2
+## TODO: switch to a release when it's out
+git+https://github.com/PaddlePaddle/PaddleNLP
```
osanseviero, line 6: Could we pin to a specific commit? That way we'll have certainty of what's released

sijunhe: this is temporary, just to get the unit test passed.
We'll do a release of PaddleNLP before we finalized this PR and lock this to a version

docker_images/paddlenlp/app/pipelines/fill_mask.py
```diff
@@ -13,7 +13,7 @@ def __call__(self, inputs: str) -> List[Dict[str, Any]]:
         Args:
             inputs (:obj:`str`): a string to be filled from, must contain the [MASK] token (check model card for exact name of the mask)
         Return:
-            A :obj:`list`:. The object returned should be like [{"entity_group": "XXX", "word": "some word", "start": 3, "end": 6, "score": 0.82}] containing :
+            A :obj:`list`:. a list of dicts containing the following:
```
osanseviero, line 16: With the restriction it always contains only one element

Narsil (approved):

PR #155, opened 2022-11-28
[PaddleNLP] Pin paddlenlp version and fix fill_mask output format

sijunhe:
- Pin paddlenlp version
- Fix fill_mask output format
- contribute `conversational` and `fill-mask` task back to `common` folder

### DIFF
--- a/docker_images/paddlenlp/requirements.txt
+++ b/docker_images/paddlenlp/requirements.txt
@@ -2,5 +2,4 @@ starlette==0.14.2
 api-inference-community==0.0.23
 huggingface_hub>=0.10.1
 paddlepaddle==2.3.2
-## TODO: switch to a release when it's out
-git+https://github.com/PaddlePaddle/PaddleNLP
+paddlenlp>=2.4.4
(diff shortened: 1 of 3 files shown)


docker_images/paddlenlp/requirements.txt
```diff
@@ -2,5 +2,4 @@ starlette==0.14.2
 api-inference-community==0.0.23
 huggingface_hub>=0.10.1
 paddlepaddle==2.3.2
-## TODO: switch to a release when it's out
-git+https://github.com/PaddlePaddle/PaddleNLP
+paddlenlp>=2.4.4
```
sijunhe, line 5: is this the right way to ensure the docker image stays in sync with the latest PaddleNLP version? We will ship more HF-compatible model and tasks in future releases.

Narsil: > is this the right way to ensure the docker image stays in sync with the latest PaddleNLP version? We will ship more HF-compatible model and tasks in future releases.

Yes it is, but it would update only when rebuilding the docker image (which would happen on source change, but not otherwise.

For reproducibility, pinning an exact version is better, for convenience this should be OK.

Narsil (approved):

PR #158, opened 2022-12-05
Add export tasks

xianbaoqian:
Context: https://huggingface.slack.com/archives/C02ERUY9TKK/p1669743642830359

### DIFF
--- a/.github/workflows/python-api-export-tasks.yaml
+++ b/.github/workflows/python-api-export-tasks.yaml
@@ -1,8 +1,8 @@
 name: Export tasks
 on:
   push:
-    paths:
-      - "**"
+    branches:
+      - main
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/python-api-export-tasks.yaml
+++ b/.github/workflows/python-api-export-tasks.yaml
@@ -1,6 +1,6 @@
 name: Export tasks
 on:
-  pull_request:
+  push:
     paths:
       - "**"
 jobs:
@@ -22,5 +22,3 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - run: python scripts/export_tasks.py
-        env: # Or as an environment variable
-          API_TOKEN: ${{ secrets.USER_API_TOKEN }}
--- a/.github/workflows/python-api-export-tasks.yaml
+++ b/.github/workflows/python-api-export-tasks.yaml
@@ -1,0 +1,26 @@
+name: Export tasks
+on:
+  pull_request:
+    paths:
+      - "**"
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8"]
+
+    steps:
+      - run: |
+          sudo apt-get update
+
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - run: python scripts/export_tasks.py
+        env: # Or as an environment variable
+          API_TOKEN: ${{ secrets.USER_API_TOKEN }}
--- a/scripts/export_tasks.py
+++ b/scripts/export_tasks.py
@@ -1,10 +1,12 @@
 """Exports a library -> supported tasks mapping in JSON format. 
 
 This script
-- parse the source code of app/main.py and extract the AST
-- find the ALLOWED_TASKS variable and get all the keys.
+- parses the source code of a library's app/main.py and extracts the AST
+- finds the ALLOWED_TASKS variable and get all the keys.
+- prints the library name as well as its tasks in JSON format.
 
-Note that transformer library is not included in the output.
+Note that the transformer library is not included in the output 
+as we can assume that it supports all tasks.
 """
 
 import ast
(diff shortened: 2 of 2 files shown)


xianbaoqian:
image

xianbaoqian:
{
    "adapter_transformers": [
        "question-answering",
        "text-classification",
        "token-classification"
    ],
    "allennlp": [
        "question-answering"
    ],
    "asteroid": [
        "audio-source-separation",
        "audio-to-audio"
    ],
    "diffusers": [
        "text-to-image"
    ],
    "doctr": [
        "object-detection"
    ],
    "espnet": [
        "text-to-speech",
        "automatic-speech-recognition"
    ],
    "fairseq": [
        "text-to-speech",
        "audio-to-audio"
    ],
    "fastai": [
        "image-classification"
    ],
    "fasttext": [
        "feature-extraction",
        "text-classification"
    ],
    "flair": [
        "token-classification"
    ],
    "k2_sherpa": [
        "automatic-speech-recognition"
    ],
    "keras": [
        "image-classification"
    ],
    "nemo": [
        "automatic-speech-recognition"
    ],
    "paddlenlp": [
        "conversational",
        "fill-mask"
    ],
    "pyannote_audio": [
        "automatic-speech-recognition"
    ],
    "sentence_transformers": [
        "feature-extraction",
        "sentence-similarity"
    ],
    "sklearn": [
        "tabular-classification",
        "tabular-regression",
        "text-classification"
    ],
    "spacy": [
        "token-classification",
        "text-classification",
        "sentence-similarity"
    ],
    "speechbrain": [
        "audio-classification",
        "audio-to-audio",
        "automatic-speech-recognition",
        "text-to-speech",
        "text2text-generation"
    ],
    "stanza": [
        "token-classification"
    ],
    "superb": [
        "automatic-speech-recognition",
        "speech-segmentation"
    ],
    "timm": [
        "image-classification"
    ]
}

.github/workflows/python-api-export-tasks.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - run: python scripts/export_tasks.py
+        env: # Or as an environment variable
+          API_TOKEN: ${{ secrets.USER_API_TOKEN }}
```
osanseviero, line 26: Is this needed?

scripts/export_tasks.py
```diff
@@ -0,0 +1,40 @@
+import ast
```
osanseviero, line 1: Maybe add a short docstring at the top explaining what this does/what's the output of the script

osanseviero: Also mention the limitation that it does not include transformers

scripts/export_tasks.py
```diff
@@ -0,0 +1,40 @@
...
+import os
+import pathlib
+import json
+
+lib_to_task_map = {}
+
+
+def extract_tasks(library_name, variable_name, value):
```
osanseviero, line 9: Nit, a good way to indicate this method is intended to be private

```suggestion
def _extract_tasks(library_name, variable_name, value):
```

osanseviero, line 9: Would be great if you could add docstrings in these methods. What is `variable_name` and what is `value`? Would also be good to add some typings

scripts/export_tasks.py
```diff
@@ -0,0 +1,40 @@
...
+    if variable_name == "ALLOWED_TASKS":
+        if isinstance(value, ast.Dict):
+            for key in value.keys:
+                lib_to_task_map.setdefault(library_name, []).append(key.value)
```
osanseviero, line 13: Rather than using `setdefault`, a nice way is to use `defaultdict` when you define `lib_to_task_map. Then you can specify a default value

```python
from collections import defaultdict
dct = defaultdict(list)
```

scripts/export_tasks.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def traverse_global_assignments(library_name, file_content, handler):
```
osanseviero, line 16: Same as above, some docstrings here would be useful

scripts/export_tasks.py
```diff
@@ -0,0 +1,40 @@
...
+    for element in ast.parse(file_content).body:
+        # Typical case
+        if isinstance(element, ast.AnnAssign):
+            handler(library_name, element.target.id, element.value)
```
osanseviero, line 20: Out of curiosity, what is `target.id` here?

.github/workflows/python-api-export-tasks.yaml
```diff
@@ -0,0 +1,26 @@
+name: Export tasks
+on:
+  pull_request:
+    paths:
+      - "**"
```
osanseviero, line 5: I would likely run this on merge to main rather than for every pull request

osanseviero (commented):
Very nice! The solution seems fine to me. I left some suggestions to make this easier to maintain and understand, but overall LG

scripts/export_tasks.py
xianbaoqian: Sure. target is name. it contains both id which is the variable name and also a context to differentiate loading, deleting and assignment. Ref: https://docs.python.org/3/library/ast.html#ast.Store

osanseviero (approved):
LGTM!

PR #162, opened 2022-12-14
Delete docker_images/superb directory

xianbaoqian:
Context: https://github.com/huggingface/hub-docs/pull/552#discussion_r1047105960

> Superb repos use the Community Inference API, but given they are not a library and their usage is more for a competition, they were not added to the Libraries.ts (confirmed with @lewtun). From a discussion, we think https://huggingface.co/models?library=superb&sort=modified are mostly old legacy models and it's ok to remove them from the Inference API, as the official ones now use transformers. (the old integration allowed users to specify their own requirements and inference code but that was deprecated)


osanseviero (approved):
Bye bye my old code :(

osanseviero:
You'll also need to modify https://github.com/huggingface/api-inference for the parts that use superb

xianbaoqian:
> You'll also need to modify https://github.com/huggingface/api-inference for the parts that use superb

Sure. Will do!

PR #163, opened 2022-12-14
Delete docker_images/doctr directory

xianbaoqian:
Context: https://github.com/huggingface/hub-docs/pull/552#discussion_r1047105960

There is only 1 model of this type.


osanseviero (changes requested):
As discussed, let's keep doctr as we now can handle libraries not displayed at left of hf.co/models

xianbaoqian:
Yes. Let's revert this one.

PR #164, opened 2022-12-14
Rename k2_sherpa to k2

xianbaoqian:
Context https://github.com/huggingface/hub-docs/pull/552#discussion_r1048464435

It's called k2_sherpa in the inference API but in the repos, people just use k2 so let's make it consistent so that the export script can work.


.github/workflows/python-api-k2.yaml
```diff
@@ -1,10 +1,10 @@
 # TODO: Merge this with allenNLP to have a single workflow for all docker images.
```
osanseviero, line 1: Would you mind adding a second workflow as this one? https://github.com/huggingface/api-inference-community/blob/main/.github/workflows/python-api-flair-cd.yml 

That will enabled continuous deployment

osanseviero (commented):
LGTM but it might require some changes in api-inference-community. 

@Narsil I don't see k2 in places such as https://github.com/huggingface/api-inference/blob/main/master/app/types.py#L21, but the API works. How come? :D

xianbaoqian:
> LGTM but it might require some changes in api-inference-community.
> 
> @Narsil I don't see k2 in places such as https://github.com/huggingface/api-inference/blob/main/master/app/types.py#L21, but the API works. How come? :D

Yeah. I'm curious on this too.

Narsil:
It's defined right there: https://github.com/huggingface/api-inference/blob/main/master-rs/src/data/hf_types.rs#L237

Narsil:
The rename should work.

Narsil:
> The rename should work.

Unfortunately the build is now broken, I don't know what maintainers have done.

Narsil (approved):
LGTM

xianbaoqian:
Thanks for fixing the bug @Narsil !

PR #173, opened 2023-01-24
Fix backwards compatibility for inference API for scikit-learn models with no config

merveenoyan:
My previous PR was raising a UserWarning which results in inference API to break when there's no config.json in scikit-learn models. (which applies to all the sklearn models not uploaded with skops) so I swapped raising `UserWarning` with `warnings.warn` instead.
*(to me it doesn't make any sense when a widget doesn't include example input anyway which doesn't come with a scikit-learn model itself but it's discussion for another day)*
Ekran Resmi 2023-01-24 16 47 54


Narsil (approved):
LGTM.

We can ignore the test failure, it's because micromamba is slow to start the docker (there's no way around it)

PR #174, opened 2023-01-24
Integrate skops serialization format

merveenoyan:


docker_images/sklearn/app/pipelines/common.py
```diff
@@ -37,21 +38,28 @@ def __init__(self, model_id: str):
...
         except Exception:
             config = dict()
-            raise UserWarning("`config.json` does not exist or is invalid.")
+            warnings.warn("`config.json` does not exist or is invalid.")
 
         self.model_file = (
             config.get("sklearn", {}).get("model", {}).get("file", DEFAULT_FILENAME)
         )
+        self.model_format = (
+            config.get("sklearn", {}).get("model_format", "pickle")
```
merveenoyan, line 47: FYI this assumes file is pickle because of how there's no skops models without config (the ones without config are to maintain vanilla sklearn models that weren't uploaded with skops)

docker_images/sklearn/tests/test_api.py
```diff
@@ -7,98 +7,143 @@
...
     "tabular-classification": [
-        "skops-tests/iris-sklearn-1.0-logistic_regression-with-config",
-        "skops-tests/iris-sklearn-1.0-logistic_regression-without-config",
-        "skops-tests/iris-sklearn-1.0-hist_gradient_boosting-with-config",
-        "skops-tests/iris-sklearn-1.0-hist_gradient_boosting-without-config",
-        "skops-tests/iris-sklearn-latest-logistic_regression-with-config",
-        "skops-tests/iris-sklearn-latest-logistic_regression-without-config",
-        "skops-tests/iris-sklearn-latest-hist_gradient_boosting-with-config",
-        "skops-tests/iris-sklearn-latest-hist_gradient_boosting-without-config",
+        "skops-tests/iris-sklearn-1.0-logistic_regression-with-config-skops",
```
merveenoyan, line 10: I added a skops test case for every pickle test case **with config**, do you think I should've reduced it? The GHActions never works on these tests anyway, so there's not much limitation except for local dev. @BenjaminBossan @adrinjalali

merveenoyan (commented):
Left some notes @adrinjalali @BenjaminBossan

Narsil (approved):

docker_images/sklearn/app/pipelines/common.py
```diff
@@ -37,21 +38,29 @@ def __init__(self, model_id: str):
...
-                if len(record) > 0:
-                    # if there's a warning while loading the model, we save it so
-                    # that it can be raised to the user when __call__ is called.
-                    self._load_warnings += record
+                if self.model_format == "pickle":
+
+                    self.model = joblib.load(
+                        open(Path(cached_folder) / self.model_file, "rb")
+                    )
+                    if len(record) > 0:
```
BenjaminBossan, line 55: Should this whole check not be outside of the `if self.model_format == "pickle":` condition? Otherwise, the warning is only recorded when the format is pickle.

docker_images/sklearn/tests/generators/generate.py
```diff
@@ -34,7 +34,7 @@
 from sklearn.model_selection import train_test_split
 from sklearn.pipeline import make_pipeline
 from sklearn.preprocessing import FunctionTransformer, StandardScaler
-from skops import hub_utils
+from skops import hub_utils, io
```
BenjaminBossan, line 37: We wanted to use `import skops.io as sio`.

docker_images/sklearn/tests/generators/generate.py
```diff
@@ -130,26 +136,35 @@ def create_repos(est_name, task_name, est, sample, version):
...
+        w_or_wo="with",
+        serialization_format=serialization_format,
     )
     push_repo(repo_name=repo_name, local_repo=local_repo)
 
-    # push WIHTOUT CONFIG
-    repo_name = REPO_NAMES[task_name].format(
-        version=version, est_name=est_name, w_or_wo="without"
-    )
+    if serialization_format == "pickle":
```
BenjaminBossan, line 146: Could you explain why this is only done for "pickle"?

docker_images/sklearn/tests/test_api.py
BenjaminBossan: It would be nice if we could eventually get the GH to work, in which case it might matter. Otherwise, I don't have any opinion on that.

BenjaminBossan (commented):
Overall, looks very good, thanks! I have some minor comments only.

merveenoyan:
@adrinjalali @BenjaminBossan all of the tests pass except for the single one with methodcaller and skops. if you think this is going to be fixed on skops side I don't think we should swap it.

merveenoyan:
Addressed comments and applied style changes.

BenjaminBossan:
> if you think this is going to be fixed on skops side I don't think we should swap it.

https://github.com/skops-dev/skops/pull/287

However, even if it gets merged, it might take a while before we have our next skops release, since we just had one very recently.

merveenoyan:
@BenjaminBossan should I merge?

Narsil:
It's not deployed yet. We're migrating some stuff to support the private hub. Should be deployed tomorrow.

PR #186, opened 2023-02-01
Adding negative_prompt to diffusers.

### LINKED ISSUE #185: Enable negative_prompt param on text-to-image pipeline
Since `StableDiffusion` and `AltDiffusion`  supports negative prompting, it would be great to have it enable on the Inference API. Specially important on [Stable Diffusion v2](https://minimaxir.com/2022/11/stable-diffusion-negative-prompt/)

Narsil:
Fixes #185

### DIFF
--- a/.github/workflows/python-api-quality.yaml
+++ b/.github/workflows/python-api-quality.yaml
@@ -16,7 +16,7 @@ jobs:
           python-version: 3.8
       - name: Install dependencies
         run: |
-          pip install black isort flake8 mypy
+          pip install black==22.3 isort flake8 mypy
       - name: Make quality
         run: |
           make quality
--- a/docker_images/diffusers/app/pipelines/text_to_image.py
+++ b/docker_images/diffusers/app/pipelines/text_to_image.py
@@ -53,7 +53,9 @@ class TextToImagePipeline(Pipeline):
                 self.ldm.scheduler.config
             )
 
-    def __call__(self, inputs: str, inference_steps=25) -> "Image.Image":
+    def __call__(
+        self, inputs: str, inference_steps=25, negative_prompt=None, **kwargs
+    ) -> "Image.Image":
         """
         Args:
             inputs (:obj:`str`):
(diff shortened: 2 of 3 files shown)


apolinario:
LGTM! The only question is: would we be able (or want to) modify the inference widget to also support negative prompts?

If not, an alternative way to do it would be to use a single field but with special characters to split what is the prompt and what is the negative prompt (e.g.: Midjourney uses `::` for multi-prompting)

Narsil:
For the widget we would need @mishig25 advice.

The widget is the same for ALL `text-to-image` models.

If you're ok with using `::` we could remove the parameter here, and simply use it here, wdyt ?

That makes everything work in the widget out-of-the-box. Just this `::` might not be intuitive nor discoverable..

apolinario:
I think we could keep both. The two inputs for the actual API and the parsing for the widget (but no need to make this distinction, just check if there's a `::` in the prompt and if so use that as a negative prompt)

I do agree with the intuitive/discoverable part tho. Dunno how to solve properly, as indeed negative prompts are almost fundamental for v2 SD models to yield good stuff

patrickvonplaten (approved):
Thanks!

radames:
thanks @Narsil , as long as the API support it we can now hack into the widget @mishig25 and add the syntax for a single input or two fields. Question @apolinario do all text-to-image models support negative prompt? since we only have on pipeline tag, we'll need to differentiate that for the widget

Narsil:
> I think we could keep both.

I strongly disagree. Keeping two things open, means it's going to be exponentially harder to support new use cases.
I tend to prefer `::` for the potential to growing complexity to lpw.

Discoverability, as long as we have a single widget for all is going to be an issue (I can't imagine lpw being solved by a widget and not using regexps within the original text).

radames:
have no strong opinions here, but I'd say the API should match the diffusers parameters, like the other APIi inference transformers endpoints. Maybe the `::` pattern should be added direct to `diffusers` WDYT @patrickvonplaten ?

Narsil:
@apolinario Are we ok with just using `::` instead of the parameter ?

Seems the most versatile approach and we've postponed merging this for way too long imo.

patrickvonplaten:
I'm not too keen on adding `::` logic to `diffusers` that's something that could be added on top of `diffusers`

Narsil:
> I'm not too keen on adding `::` logic to `diffusers` that's something that could be added on top of `diffusers`

That's 100% understandable, but since this is visible as the API, it might come off as expected. 
Otherwise let's merge as-is, never implement `::` and just hopefully never have to merge lpw on the widgets.

PR #220, opened 2023-03-09
Add fasttext language identification

sheonhan:
Add `language-identification` for fasttext model. 

Related: https://github.com/huggingface/hub-docs/pull/703 
(cc @osanseviero)

PR #222, opened 2023-03-13
fastText language identification follow up

sheonhan:
Make some missing changes from https://github.com/huggingface/api-inference-community/pull/220


sheonhan:
Now it makes sense why not adding the "text-classification" tag caused the widget to not work properly. With that, it works! 
Screenshot_2023-03-13_13 57 58

docker_images/fasttext/tests/test_api_text_classification.py
```diff
@@ -87,10 +91,11 @@ def test_malformed_question(self):
 
     def test_multiple_words(self):
         inputs = "this is great"
-
+        
+        expected_status_code = 200 if "language-identification" in self.model_id else 400
```
osanseviero, line 95: Maybe add a docstring explaining why this is the case

osanseviero (approved):
Thanks a lot! :fire:

Narsil:
For the quality it's still `pip install -e .[quality] &&  make style`. Thanks

docker_images/fasttext/tests/test_api_text_classification.py
sheonhan: I added a comment why this is the case (since I didn't see docstring used elsewhere) but let me know if you meant docstring!

PR #228, opened 2023-04-11
Change input format on call for better compatibility with kwargs

radames:
As we always have a batch size of 1 for a single prompt, let's improve compatibility with kwargs. For instance, use:
```
inputs: "...prompt",
parameters {
  negative_prompt: "....negative_prompt"
}
```
instead of:
```
inputs: "...prompt",
parameters {
  negative_prompt: ["....negative_prompt"]
}
```

currently if you don't send negative_prompt as a list you get this error

```
{"error":"`negative_prompt` should be the same type to `prompt`, but got <class 'str'> != <class 'list'>."}
```


Narsil (approved):
LGTM !

PR #230, opened 2023-04-13
Adding `image-to-image` + `zero-shot-image-classification` tasks.

Narsil:


api_inference_community/validation.py
```diff
@@ -337,6 +340,15 @@ def ffmpeg_read(bpayload: bytes, sampling_rate: int) -> np.array:
 def normalize_payload_image(bpayload: bytes) -> Tuple[Any, Dict]:
     from PIL import Image
 
+    try:
```
osanseviero, line 343: Maybe add a short docstring to explain which is the use case for this handling

osanseviero (approved):
Thanks!

PR #232, opened 2023-04-18
Initial open_clip zero-shot support. Test working locally but docker failing.

rwightman:


Narsil (approved):
LGTM. Everything is in order.

On the hub we're going to have issues the widget is sending data a bit different than what you're (correctly) expecting.

We need to `candidate_labels.split(',')` so users can send comma separated labels (this is what the hub widget does). We need to update `api-inference-community` to allow `image` key to contain the image (isntead of `inputs`).

I have done both changes in a separate branch, if you're OK I could push here directly.

https://github.com/huggingface/api-inference-community/pull/233

PR #235, opened 2023-04-18
Upgradign sentence transformers (huggingface_hub version).

Narsil:
Should fix https://huggingface.co/sentence-transformers/paraphrase-xlm-r-multilingual-v1

Which is currently bugged (it seems that the old huggingface_hub is not properly respecting the HEAD/GET requests with tokens which are necessary for sentence-transformers, leading to BAD requests)


osanseviero (approved):
Thanks! :fire:

PR #241, opened 2023-04-28
Diffusers image updates

oOraph:
- Enable HF transfer
- No more prestart script systematically run, means one less model load
- Logging when running prestart by hand


Narsil:
LGTM.

@osanseviero  @patrickvonplaten just FYI. We added some more logging as we're trying to enable faster load times and better GPU sharing (to be able to load more models on the same compute)

patrickvonplaten:
Good for me!

Narsil (approved):
LGTM

PR #243, opened 2023-05-02
set is_controlnet default value

radames:
when model doesn't have a config file `is_controlnet` has no default value


patrickvonplaten (approved):
Thanks!

Narsil (approved):
LGTM

PR #244, opened 2023-05-02
rename img2img api params

radames:
rename api as discussed.
couldn't get the tests to run well without running with a GPU docker backend, is it normal?

### DIFF
--- a/docker_images/diffusers/app/pipelines/image_to_image.py
+++ b/docker_images/diffusers/app/pipelines/image_to_image.py
@@ -72,12 +72,14 @@ class ImageToImagePipeline(Pipeline):
                 self.ldm.scheduler.config
             )
 
-    def __call__(self, image: Image.Image, inputs: str = "", **kwargs) -> "Image.Image":
+    def __call__(
+        self, inputs: Image.Image, prompt: str = "", **kwargs
+    ) -> "Image.Image":
         """
         Args:
-            inputs (:obj:`str`):
+            prompt (:obj:`str`):
                 a string containing some text
-            image (:obj:`PIL.Image.Image`):
+            inputs (:obj:`PIL.Image.Image`):
                 a condition image
         Return:
             A :obj:`PIL.Image.Image` with the raw image representation as PIL.
@@ -94,10 +96,10 @@ class ImageToImagePipeline(Pipeline):
             if "num_inference_steps" not in kwargs:
                 kwargs["num_inference_steps"] = 25
             images = self.ldm(
-                inputs,
-                image,
+                prompt,
+                image=inputs,
                 **kwargs,
             )["images"]
         else:
-            images = self.ldm(inputs, image, **kwargs)["images"]
+            images = self.ldm(prompt, image=inputs, **kwargs)["images"]
         return images[0]
--- a/docker_images/diffusers/app/pipelines/image_to_image.py
+++ b/docker_images/diffusers/app/pipelines/image_to_image.py
@@ -72,14 +72,12 @@ class ImageToImagePipeline(Pipeline):
                 self.ldm.scheduler.config
             )
 
-    def __call__(
-        self, inputs: Image.Image, prompt: str = "", **kwargs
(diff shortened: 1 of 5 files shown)


Narsil:
Thank you for this.

Pinging @mishig25 for the update on the widget.

For code quality `make style` should fix it (It's mostly the same rules as transformers, just slightly outdated)

patrickvonplaten (approved):
Ok for me!

radames:
ok, I've increased `timeout=600` for the `hf-internal-testing/tiny-controlnet` test and it passes locally, however it's failing on the CI 🤔 wdty @Narsil ?

docker_images/diffusers/app/pipelines/image_to_image.py
```diff
@@ -94,10 +96,10 @@ def __call__(self, image: Image.Image, inputs: str = "", **kwargs) -> "Image.Ima
             if "num_inference_steps" not in kwargs:
                 kwargs["num_inference_steps"] = 25
             images = self.ldm(
-                inputs,
-                image,
+                prompt,
+                image=inputs,
```
Narsil, line 100: This make me double `s/image/inputs` is necessary.

We could by all means keep the name `image` here no ? (Since the validation accepts both and sends them as a positional argument).

Works for me either way, but renaming in such way is always something I'm careful about.

radames: ow, I see what you mean, since the the position matches the image? that totally works ☺️ right?
`self.ldm(prompt, inputs, **kwargs)["images"]`

radames: or did you mean, not rename the `image` to `inputs`?

`def __call__(self, image: Image.Image, prompt: str = "", **kwargs) -> "Image.Image":`

radames: btw,, if you or @mishig25 would like to update/merge this based on your feedback tomorrow morning, please feel free to do so. We don't need to delay this, especially considering the few dependent issues 🙏

Narsil: I meant not reanme `image` to `inputs` (in this particular piece of code, the widget and the API should still handle both)

radames: ok @Narsil that makes sense! I've update the code

radames:
@Narsil still not clean how to run these [tests](https://github.com/huggingface/api-inference-community/tree/main/docker_images/diffusers/tests)
when I do `pytest -sv --rootdir docker_images/diffusers docker_images/diffusers` the  `TESTABLE_MODELS` are not loaded via the docker container.

Narsil:
> @Narsil still not clear how to run these [tests](https://github.com/huggingface/api-inference-community/tree/main/docker_images/diffusers/tests)

These are local tests.

The docker tests are run like so

```
RUN_DOCKER_TESTS=1 pytest -sv tests/test_dockers.py::DockerImageTests::test_diffusers
```

You can also test dynamically more practically imo:

```
./manage.py docker hf-internal-testing/tiny-controlnet
```

This will start the docker as the API would and then you can query it classically.
It's running the full docker environment.

If you want dynamic debugging (like pdb shell)
```
./manage.py start hf-internal-testing/tiny-controlnet
```
And this will start with your local env, meaning you can do all sorts of interactions.

Narsil (approved):
LGTM

Narsil:
I'll go aahead and merge the model seems to be working correctly on docker.

radames:
hi @Narsil is it already deploy to pro?

Narsil:
Yes it's autodeployed normally.

radames:
yes, however sending requests I still get this error ` Error: __call__() got multiple values for argument 'prompt'` 
testing locally docker, it works as expected 🤔

Narsil:
Which model ? Old images that are running are not flushed.

PR #246, opened 2023-05-06
Updade xformers

radames:
found another issue
using an olde formers version
no need to use the prebuild since xformers is now available on pypi


patrickvonplaten (approved):
Perfect! Actually ideal would be to always install PyTorch 2.0, then we wouldn't have to rely on xformers anymore at all

Narsil (approved):
LGTM Thanks !

PR #247, opened 2023-05-08
Add support for neural compressor models

echarlaix:


docker_images/neural_compressor/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM tiangolo/uvicorn-gunicorn:python3.8
+LABEL maintainer="Ella Charlaix <ella@huggingface.co>"
+
+RUN apt-get update -y && apt-get install ffmpeg libsm6 libxext6 -y
+RUN pip install --no-cache-dir torch==1.13.0 torchvision==0.14.0 torchaudio==0.13.0
```
Narsil, line 5: Should we maybe update to 2.0 ?

Narsil (commented):
Awesome PR.

neural_compressor is going to become a `framework` right ? (`library_tag` in the README) .

@osanseviero for confirmation about this.

After this is merge, just need to add the hub thing for frameworks + API inference variant.

docker_images/neural_compressor/Dockerfile
echarlaix: Good point, will do!

echarlaix:
> neural_compressor is going to become a `framework` right ? (`library_tag` in the README) .

Makes sense to me ! An other option could be to have `optimum` as `library_tag`, and have `neural-compressor` / `openvino` / `onnx` as tags, if needed we can also add the model loading logic to `optimum` (mentionning ONNX and OpenVINO as I would like to add their support as well)

What do you think would make more sense @Narsil @osanseviero ?

docker_images/nemo/app/pipelines/automatic_speech_recognition.py
```diff
@@ -58,7 +58,7 @@ def __call__(self, inputs: np.array) -> Dict[str, str]:
             transcriptions = self.model.transcribe([audio_path])
 
             # if transcriptions form a tuple (from RNNT), extract just "best" hypothesis
-            if type(transcriptions) == tuple and len(transcriptions) == 2:
+            if type(transcriptions) is tuple and len(transcriptions) == 2:
```
echarlaix, line 61: Modified so that the code quality test pass (needed for `flake8` latest version / `pycodestyle>=`[v2.11.0](https://github.com/PyCQA/pycodestyle/blob/2.11.0/pycodestyle.py#L1452))

Narsil:
> What do you think would make more sense @Narsil @osanseviero ?

Both are fine to me. I don't necessarily know the scoping of this vs optimum and such.

echarlaix:
Perfect thanks @Narsil, I need to wait for updates from the Intel collaboration before merging, will change the PR status to draft temporarily

PR #248, opened 2023-05-10
add library `mmclassification`

alibool:


osanseviero:
Hi there! It seems there are no models on the Hub tagged with this library (https://huggingface.co/models?library=mmclassification). In general, we expect there to be few dozens of models of a library before adding the API and the filters at the left of hf.co/models.

osanseviero:
Closing for now, but feel free to reopen if there are more models!

PR #275, opened 2023-05-17
Periodically unload model / reset worker if no request was received for a while

oOraph:
Allows to free gpu resources for example and the cost to move it back to gpu if already loaded in mem is only ~3s. 

Add some overcommit possibility when serving several models on a single gpu at once

tl;dr

The Api Inference infra now serves most diffusion models on shared gpus: thus we can simultaneously serve the very same number of diffusion models as before with twice less hosts dedicated for this. However, we could improve the stacking models on a single gpu by adding some overcommit possibility. This is the point of the pr: releasing the gpu resources when idle, more precisely the gpu memory that is currently the main factor preventing us from increasing the model stacking ratio on a gpu. In this pr the Guvicorn worker is simply killed as soon as we see that the model did not receive any request for a while (but the asgi server respawns one immediately so that's ok). This releases the gpu resources and allows for other models to concurrently serve requests. Meanwhile the model is still loaded in ram+swap, ready to be migrated on gpu in ~2-4 seconds at the next request. WIthout this, some gpu resources stay allocated for this model as long as the asgi worker runs.

### DIFF
--- a/docker_images/diffusers/app/pipelines/image_to_image.py
+++ b/docker_images/diffusers/app/pipelines/image_to_image.py
@@ -2,6 +2,7 @@ import json
 import os
 
 import torch
+from app import idle, timing
 from app.pipelines import Pipeline
 from diffusers import (
     AltDiffusionImg2ImgPipeline,
@@ -100,10 +101,6 @@ class ImageToImagePipeline(Pipeline):
         else:
             raise ValueError("Model type not found or pipeline not implemented")
 
-        if torch.cuda.is_available():
-            self.ldm.to("cuda")
-            self.ldm.unet.set_attn_processor(AttnProcessor2_0())
-
         if isinstance(
             self.ldm,
             (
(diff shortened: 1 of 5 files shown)


docker_images/diffusers/app/pipelines/text_to_image.py
```diff
@@ -38,9 +44,6 @@ def __init__(self, model_id: str):
         self.ldm = DiffusionPipeline.from_pretrained(
             model_to_load, use_auth_token=os.getenv("HF_API_TOKEN"), **kwargs
         )
-        if torch.cuda.is_available():
```
oOraph, line 41: @pcuenca, @patrickvonplaten @radames I would need your advice here: is there any requirement for the previous specif order of operations ? I mean: does to('cuda') need to be called before loading lora weights and assigning the scheduler or it this current proposal ok ?

docker_images/diffusers/app/pipelines/image_to_image.py
```diff
@@ -94,10 +96,6 @@ def __init__(self, model_id: str):
         else:
             raise ValueError("Model type not found or pipeline not implemented")
 
-        if torch.cuda.is_available():
```
oOraph, line 97: @pcuenca, @patrickvonplaten @radames I would need your advice here too. see comment:
https://github.com/huggingface/api-inference-community/pull/275/files#r1196745186

radames (commented):
~hi @oOraph , if I'm not wrong, and @Narsil can confirm, we have an internal deployment that takes care of the unloading and loading of images from the api-inference-community and it does exactly what you're proposing. Thank you for the PR anyways.~

Narsil (approved):
LGTM.

We need to test in prod in collaboration the scheduler to make sure it works properly anyway :)

docker_images/diffusers/app/pipelines/text_to_image.py
patrickvonplaten: Both work :-)

PR #286, opened 2023-06-01
run style on diffusers app

davanstrien:
Currently, `docker_images/diffusers/app/idle.py` is causing tests to fail. This PR runs formatting to fix this file.


Narsil:
Thanks @davanstrien

PR #289, opened 2023-06-12
Creating dummy change for `nemo` to upgrade version.

Narsil:


Vaibhavs10 (approved):
Thank you for helping with this @Narsil! A follow-up question, just so that I understand this correctly, for every update that `nemo` has we'd have to create a dummy PR to update the `requirements.txt` file right?
Or we can just switch to a version on PyPI and then bump up the version with every release too.

Narsil:
Yup, either way, fixing the pypi version makes it more readable what the intent is, but yes, there is no auto deploy otherwise.

If it becomes ultra tedious we can maybe think of something, but in the meantime let's keep it simple if you're ok.

PR #292, opened 2023-06-13
Update NeMo version

titu1994:
PR to update the NeMo version to support Fast Conformer API Inference (+ switch to smaller model for CI)

cc @Vaibhavs10


Narsil (approved):
LGTM

PR #299, opened 2023-06-23
Added missing test to PEFT tests

merveenoyan:
this was the reason why docker tests weren't working @Narsil

### DIFF
--- a/docker_images/peft/tests/test_api.py
+++ b/docker_images/peft/tests/test_api.py
@@ -1,0 +1,61 @@
+import os
+from typing import Dict
+from unittest import TestCase, skipIf
+
+from app.main import ALLOWED_TASKS, get_pipeline
+
+
+# Must contain at least one example of each implemented pipeline
+# Tests do not check the actual values of the model output, so small dummy
+# models are recommended for faster tests.
+TESTABLE_MODELS: Dict[str, str] = {
+    # IMPLEMENT_THIS
+    # "automatic-speech-recognition": "mysample-ASR",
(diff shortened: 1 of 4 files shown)


merveenoyan:
@Narsil the tests time out because the model is not taken to GPU in docker tests, do you have any suggestion?

Narsil:
We have no GPU on the CI.
the models won't run on GPU either on the cluster.

merveenoyan:
@Narsil do you have any suggestion? e.g. what are you doing with diffusers tests? this is a relatively small model that I'm testing with

Narsil:
Can't it work on CPU ? the test should work on CPU.

osanseviero:
To echo @Narsil, the test models should be small and run on CPU. If the code is failing because no GPU is available, then there's a bug in the code and it should be adapted to make sure it runs without GPU.

docker_images/peft/tests/test_api.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import Dict
+from unittest import TestCase, skipIf
+
+from app.main import ALLOWED_TASKS, get_pipeline
+
+
+# Must contain at least one example of each implemented pipeline
+# Tests do not check the actual values of the model output, so small dummy
+# models are recommended for faster tests.
+TESTABLE_MODELS: Dict[str, str] = {
```
osanseviero, line 11: You also need to have a model here, otherwise `docker_images/peft/tests/test_api_text_generation.py` will never run

PR #308, opened 2023-07-25
Update to SpeechBrain 0.5.15

TParcollet:
Hi there! We are bringing some new interface with this update. I have tested the new ones and everything works (with your test suit) except one: speechbrain/tts-fastspeech2-ljspeech It returns a wrong code, but I don't understand why and can't find it on the log, maybe you could help? I've tested the interface locally and it works (if we follow the instruction on the card) that are similar to the one I've put on the pipeline :-( 

Thank you folks!


osanseviero:
The failing error is a docker test, so it's not the same as the ones you were testing locally. It's likely some timeout issue unrelated to this PR though. I see from history that this workflow is no working very well (https://github.com/huggingface/api-inference-community/actions/workflows/python-api-speechbrain.yaml)

@Vaibhavs10, would appreciate a review here

Vaibhavs10:
Just looked through this @Narsil it looks like a pydantic-V2 error `pydantic.errors.PydanticImportError: `pydantic:ConstrainedFloat` has been removed in V2.` this is causing the image build issues AFAICT.

However, as I can see that we've already made a patch to bump to pydantic v2 2 weeks ago.

Vaibhavs10:
Ah. I think I found the culprit - we need to bump this to `0.0.31` - https://github.com/huggingface/api-inference-community/blob/186c2029237f871ac202aa4ca100149dc3ac0307/docker_images/speechbrain/requirements.txt#L3

Vaibhavs10 (approved):
The pipelines look good, tests look good too. Thanks for adding this @TParcollet, the only change left would be to bump up the version of `api-inference-community` in the `requirements.txt`

Vaibhavs10:
Not sure why it is still failing. @Narsil - Can you please take a look as well? I don't see any obvious issues with this.

Narsil:
We're on version 0.0.32 https://pypi.org/project/api-inference-community/

Vaibhavs10:
Hmm. I bumped the version and it seems to have fixed the `pydantic` validation error. But now the [docker tests](https://github.com/huggingface/api-inference-community/actions/runs/5672252677/job/15371086151?pr=308) are failing: https://github.com/huggingface/api-inference-community/blob/186c2029237f871ac202aa4ca100149dc3ac0307/tests/test_dockers.py#L649

The test for malformed audio file test throws a `500`. This looks like a mismatch between the expected input vs. input provided. Can you please investigate this @TParcollet 🙏

osanseviero:
This has been broken for...ever it seems. https://github.com/huggingface/api-inference-community/actions/workflows/python-api-speechbrain.yaml

I think the error is unrelated to this PR, so I would be ok going forward with this one

Vaibhavs10:
Okay, perfect! Going to merge this then! 🤗

PR #309, opened 2023-07-26
Fixes a race condition on api inference config update

oOraph:


Narsil (approved):
LGTM so you can test when you're ready

PR #313, opened 2023-07-27
add SpeechBrain Speech-Emotion-Diarization interface

BenoitWang:
Hi, 

Following #308, this PR adds support for the speech-emotion-diarization model. This `audio_diarization` interface can benefit other diarization tasks (emotion/speaker/vad/other speech events).

For more details:
[speechbrain/emotion-diarization-wavlm-large](https://huggingface.co/speechbrain/emotion-diarization-wavlm-large) 

Thanks!

### DIFF
--- a/docker_images/speechbrain/app/main.py
+++ b/docker_images/speechbrain/app/main.py
@@ -6,6 +6,7 @@ from typing import Dict, Type
 from api_inference_community.routes import pipeline_route, status_ok
 from app.pipelines import (
     AudioClassificationPipeline,
+    AudioDiarizationPipeline,
     AudioToAudioPipeline,
     AutomaticSpeechRecognitionPipeline,
     Pipeline,
@@ -39,6 +40,7 @@ logger = logging.getLogger(__name__)
 # directories. Implement directly within the directories.
 ALLOWED_TASKS: Dict[str, Type[Pipeline]] = {
     "audio-classification": AudioClassificationPipeline,
+    "audio-diarization": AudioDiarizationPipeline,
     "audio-to-audio": AudioToAudioPipeline,
     "automatic-speech-recognition": AutomaticSpeechRecognitionPipeline,
     "text-to-speech": TextToSpeechPipeline,
--- a/docker_images/speechbrain/app/common.py
+++ b/docker_images/speechbrain/app/common.py
@@ -20,6 +20,8 @@ class ModelType(Enum):
     FASTSPEECH2 = "FASTSPEECH2"
     # text2text-generation
     GRAPHEMETOPHONEME = "GRAPHEMETOPHONEME"
+    # audio-diarization
+    SPEECHEMOTIONDIARIZATION = "SPEECHEMOTIONDIARIZATION"
 
 
 def get_type(model_id, interface_type="interface"):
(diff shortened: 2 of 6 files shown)


docker_images/speechbrain/app/main.py
```diff
@@ -39,6 +40,7 @@
 # directories. Implement directly within the directories.
 ALLOWED_TASKS: Dict[str, Type[Pipeline]] = {
     "audio-classification": AudioClassificationPipeline,
+    "audio-diarization": AudioDiarizationPipeline,
```
Narsil, line 43: It seems that the model being tested is under `audio-classification`.

As far as I'm aware, there's no `audio-diarization` task on the hub.
Any reason to use a different name ?

BenoitWang: Hi @Narsil, though I put it under audio-classification, these tasks (e.g. speaker-diarization, voice activity detection, emotion-diarization) are not really audio-classification tasks. They are more like audio activity/event detection. Given an audio, the output for these tasks should be: which event happens from which time to which time. 
What would you suggest please?

Narsil: Looks like audio-segmentation.

Don't think we have any model under that category I think.
@osanseviero wdyt ?

BenoitWang: a similar task I find is `voice-activity-detection`, but VAD particularly refers to voice/unvoice detection. There are some diarization models under this task, maybe it's better to enlarge it to `audio-activity-detection`, like this all the vad models and diarization models can be put under it.

BenoitWang: Hi @osanseviero , what do you think of the naming please?

osanseviero: I think this makes sense. We could make `voice-activity-detection` a `subtasks` of `audio-activity-detection` if it makes sense. More generally, we're having some discussions around `text-to-speech` and `text-to-audio`, etc. Voice activity detection is likely more searched than `audio-activity-detection`. 

My suggestion would be
1. For this PR, let's use `voice-activity-detection` and make sure things work. That way we don't need to add a new type right away.
2. We need to create a new widget for diarization (cc @mishig25 )
3. Once things work, we can migrate from the more granular task to the generalization

WDYT?

BenoitWang: Thanks for the reply, I can change it to `voice-activity-detection` for now. A short summary of the two options:
1. make `voice-activity-detection` a subtask of `audio-activity-detection`
2. since `voice-activity-detection` is more searched, we don't touch it and simply add `audio-diarization` for all the diarization tasks

And yes I find speech tasks are less distinguished than NLP/CV, it would be great if that can be improved. For example TTS and TTA are two different tasks for me, even though they are both generative tasks, TTS demands a strict alignment between synthesized audio and the text (TTA  not necessarily), while TTA is based on a less strict condition and learns embeddings of various compositions. Thus the objective of the two tasks and the model design can be very different.
What do you think please? @osanseviero @mravanelli @TParcollet

Narsil: Currently we think of tasks as Input/Output not semantics nor implementation detail.

This makes reasoning from non ML perspective easier. Having additionals tags `subtasks` is great for keeping it discoverable/clean for ML people.

PR #334, opened 2023-09-26
LoRA mutualization: keep base model loaded, load/unload LoRA adapters on the fly

oOraph:
Depends on https://github.com/huggingface/api-inference-community/pull/335

PR #335, opened 2023-09-27
Update api_inference_community lib to support a new header

oOraph:
This intermediate pr is required to release a new lib version (0.0.33) that would then be used in the following pr:
https://github.com/huggingface/api-inference-community/pull/334


Narsil (approved):
LGTM

PR #339, opened 2023-10-03
feat diffusers docker image: possibility to spawn a side server reporting whether the inference service is busy or not

oOraph:
side service that will be used to serve some metrics used by custom hpa. open to discussion, not sure we actually want to have it in the following repo but we may prefer moving it in a dependency lib instead


Narsil (approved):
Nice workaround of the poor threading performance  of Python :).

Honestly I'd think the webserver should handle the backpressure better, but it's really tricky to get it right from within Python.

PR #340, opened 2023-10-05
Remove fuse/unfuse

oOraph:
Root cause not yet precisely identified but this avoids expanding some issues between adapters

related to https://github.com/huggingface/diffusers/issues/5313


Narsil (approved):
LGTM

PR #350, opened 2023-11-09
Add new library: `hezar`

arxyzan:
## What is Hezar?
This PR adds support for the `hezar` library. This library is an all-in-one AI library and framework for the Persian language and currently supports the following tasks:
- Text classification
- Token classification
- Speech recognition
- Language modeling
- Fill mask
- Image to text (OCR, Image Captioning, etc)
- Text generation

This library was officially introduced about a month ago and have been downloaded 20k+ times on PyPi already. 
Learn more about the Hezar project at [https://github.com/hezarai/hezar](https://github.com/hezarai/hezar).

## What does this PR do?
Following all the instructions in the README of this repo, we first added pipelines for the following tasks: `text-classification`, `text-generation`, `fill-mask`, `token-classification`, `automatic-speech-recognition`, `image-to-text` and then provided tests for each of them according to the provided test samples in the `common` folder and after a couple of runs, we managed to make all tests pass. 

@Narsil @osanseviero


arxyzan:
Two questions:
1. According to [the official guide](https://huggingface.co/docs/hub/models-adding-libraries#register-your-libraries-supported-tasks-on-the-hub) is it needed to also register the library [here](https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/LibrariesToTasks.ts)?
2. Is it needed to add the workflow file?

@Narsil

arxyzan:
Hey there, @Narsil @osanseviero
Is the review process going on already?

arxyzan:
Hey there @Narsil @osanseviero 
I wonder if the review is in progress since the PR was sent 2 weeks ago.
Thanks in advance.

Narsil:
Hi @arxyzan ,

Sorry for the late reply, we're all quite busy.
We're adding libraries here that have a specific tag in here : https://huggingface.co/models which afaik is not yet the case.

Pardon the curiosity, but why create an entire specific library instead of leveraging existing librairies like `transformers`. AFAIK there's nothing preventing existing architecture to run on persian (aside training/fine-tuning existing models because of the existential bias towards english in the ML community).

We have run events in the past to promote non english models for instance: https://huggingface.co/models?search=wav2vec

arxyzan:
Hi @Narsil and thanks for the response!

Regarding the specific tags, I just checked some libraries model cards from similar libraries like FlairNLP, AllenNLP, stanza, etc. and noticed that they use their library name and task under `tags`. Does this mean that we have to add tags to our model cards to make this library registerable?

Regarding your question, I understand you wondering that.
The 1st reason is that the Persian language and more specifically the Persian ML community, have always suffered from lack of hadware resources and proper datasets, hence requiring heavily optimized or small efficient models to be able to actually leverage ML models in production. Even large companies have trouble making use of BERT-like models in production which makes it hard to use models in Transformers, Diffusers, etc. (which are awesome btw!). Although we support those kind of models in Hezar, our primary goal is to provide small and efficient models that can actually be used in real world like our CRNN family of models for OCR. We have big plans for extending this library by adding such models that are not present in other HF-backed libraries. (We've just started this path and most of the current models on our Hub are for ensuring that this thing works in all scenarios).
The 2nd reason is that we took a different approach in designing pipelines for loading models, preprocessors, etc. from the Hub and the way we run end-to-end inference. If you browse the files in any of our models on the Hub you can notice it. 
The 3rd reason is that Hezar has a low-code framework structure that makes it possible to leverage Hezar's features with minor changes to your existing PyTorch model codes (See [here](https://hezarai.github.io/hezar/guide/models_advanced.html#building-models) for a showcase). Although it's kind of possible to use Transformers as a framework (e.g by subclassing PretrainedConfig and PretrainedModel), but our way of doing stuff has so much difference that it doesn't seem like a good idea to build such library on top of Transformers. 

And finally the reason for requiring this library to be registered in the HF Hub, is that
1. Users can easily test models right from the Hub 
2. The `</> Use in Hezar` tag on the top right would assist the users to load the models easily.
3. (Not sure!) The downloads are actually counted and shown on the Hub.

Overall, Hezar can be treated like other community-backed libraries that provide customized models like Flair, AllenNLP, etc. using existing or novel techniques through their own framework. This feature will really help our library and community grow and build better models and tools over time. 

Thanks for putting your time reading this 🙏🏻
Looking forward to your response 🤗

PR #352, opened 2023-11-17
Update requirements.txt

StephenHodgson:
update generic tasks to use latest version of api_inference_community


osanseviero (approved):
Thanks!

osanseviero:
I think there's no need to update this. `generic` is not a template but a quite old system. Updating the version required changing the code as well, but there's no need for this atm

PR #354, opened 2023-11-27
Bumping huggingface-hub and accelerate for PEFT image dependencies

merveenoyan:
The existing dependencies result in dependency conflict when starting the app locally.

### DIFF
--- a/docker_images/peft/requirements.txt
+++ b/docker_images/peft/requirements.txt
@@ -1,10 +1,10 @@
 starlette==0.27.0
 api-inference-community==0.0.31
-huggingface_hub==0.14.1
+huggingface_hub==0.18.0
 safetensors==0.3.1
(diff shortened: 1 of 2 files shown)


merveenoyan:
Moreover, I debugged the PEFT image locally because the widgets return `'mistral'`.
To fix it, I had to pin the transformers version because trying to load a model returns errors like `KeyError: 'mistral'` which is mapped to this: https://github.com/huggingface/transformers/issues/26759

osanseviero (approved):
Thank you! 🔥 Should we also pin peft to a specific release or commit? Installing from main can lead to undesired outcomes and makes it harder to debug any issues

merveenoyan:
@osanseviero great point! I pinned it to the latest stable release on pypi 😊

PR #355, opened 2023-11-28
👷 Automatically dispatch tasks to hf.js

coyotte508:
This will send a dispatch even to hf.js, where we need to create a matching action to receive the event with `repository_dispatch`: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch

This will allow us to automate updating https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/library-to-tasks.ts (with an action that will open a PR to this file)

cc @pcuenca @mishig25 @osanseviero @julien-c 

cc @Wauplin in case you might want that too for some reason

## Migration

Set a `REPO_DISPATCH_TOKEN` secret, a PAT that has read/write & metadata on `huggingface/huggingface.js`

### DIFF
--- a/.github/workflows/python-api-export-tasks.yaml
+++ b/.github/workflows/python-api-export-tasks.yaml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - transmit-payload-to-hfjs
 jobs:
   build:
     runs-on: ubuntu-latest


.github/workflows/python-api-export-tasks.yaml
```diff
@@ -21,4 +22,23 @@ jobs:
...
+      # Dispatch to huggingface/huggingface.js repo
+      - name: Dispatch to huggingface/huggingface.js
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.REPO_DISPATCH_TOKEN }}
+          repository: huggingface/huggingface.js
+          event-type: export-tasks
+          client-payload: |
+            {
+              "tasks": ${{ steps.export_tasks.outputs.payload }}
```
pcuenca, line 43: Nice! How is this payload received and interpreted in the remote repo, is that already in place?

julien-c (commented):
@osanseviero are we sure this is needed?

I feel like with the Inference revamp will we even still have this repo?

glegendre01:
@coyotte508, secret created. can you test it ?

osanseviero:
@julien-c fair point; let's pause this PR as a bunch of this might be unneeded in a couple of weeks and this is mostly to update a `.ts` file that does not change very often (just when a new task is added in this API)

glegendre01:
ok, keep me informed if we need to clean up the token finally

PR #362, opened 2023-12-07
Set loaded tokens to zero after removing

apolinario:


oOraph (approved):
minor:
this line 
https://github.com/huggingface/api-inference-community/pull/362/files#diff-7eb6dccb37d3ef9eabce64891b20f8ddd5a4bada2829ddbc4c7366dfff97125fR202

can now be removed (but harmless if we keep it :) )

PR #367, opened 2023-12-20
Reset default scheduler when none is specified

oOraph:


docker_images/diffusers/app/pipelines/text_to_image.py
```diff
@@ -136,6 +138,11 @@ def __call__(self, inputs: str, **kwargs) -> "Image.Image":
                 self.ldm.scheduler = SchedulerClass.from_config(
                     self.ldm.scheduler.config
                 )
+            else:
```
apolinario, line 141: Added this for the cases the scheduler isn't compatible

apolinario (approved):
LGTM!

PR #371, opened 2023-12-21
model info that works offline

oOraph:
could be used to fix https://github.com/huggingface/api-inference-community/issues/372


Narsil (approved):
LGTM but isn't there something easier with huggingface_hub ? @Wauplin maybe ?

PR #410, opened 2024-02-06
Improve manage by making it more agnostic to call method

Narsil:


tomaarsen (approved):
Works locally for me with `python manage.py` 👍

PR #415, opened 2024-02-28
Update SpeechBrain requirements.txt

Adel-Moumen:
Hello,

This PR aims at updating SpeechBrain requirements. Indeed, some HF models on our hub are using KenLM for n-gram rescoring. Hence, they require adding `pygtrie` and `kenlm` as part of the requirements in order to be used (note: these packages are not part of SpeechBrain requirements as they are optional).


TParcollet:
Simple change, not breaking, lgtm.

TParcollet (approved):
LGTM

PR #416, opened 2024-02-29
`model.config` handle breaking change in hub

julien-c:

PR #421, opened 2024-03-29
Fix broken SpeechBrain API

osanseviero:
Related to breaking change https://github.com/huggingface/api-inference-community/pull/416


julien-c (approved):
thanks for fixing

PR #424, opened 2024-05-22
Adding support for edsnlp library

percevalw:
Hi,

This PR adds support for [edsnlp](https://github.com/aphp/edsnlp), an NLP library developed at the Greater Paris University Hospitals (AP-HP) with a specific focus on hybrid models (rule-based + ML) and multi-tasking, widely used in the structuring of clinical language at scale (NER, entity linking, document classification, event extraction, ...).
Currently, we don't have a public model (except for a lightweight test model [AP-HP/dummy-ner](https://huggingface.co/AP-HP/dummy-ner/) on the hub), but we plan to release some of our non-sensitive models in the coming weeks. Would this PR be acceptable as it stands ?

Let me know if any changes are needed, of if you need more information !

### DIFF
--- a/.github/workflows/python-api-edsnlp-cd.yaml
+++ b/.github/workflows/python-api-edsnlp-cd.yaml
@@ -1,0 +1,48 @@
+name: edsnlp-docker-cd
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "docker_images/edsnlp/**"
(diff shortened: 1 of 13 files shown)


percevalw:
Closing, following the discussions on https://github.com/huggingface/huggingface.js/pull/696

PR #432, opened 2024-06-25
[`sentence-transformers`] Add sentencepiece dependency for running models with slow tokenizers only

tomaarsen:
Hello!

## Pull Request overview
* Add sentencepiece dependency for running models with slow tokenizers only

## Details
See e.g. https://huggingface.co/rokn/slovlo-v1:
image

`sentencepiece` was removed as a required dependency, because most models don't require it anymore. Nowadays, we simply throw an error if it's required but not installed. However, the API Inference should have it installed to ensure that we can also run these models with slow tokenizers. I've verified locally that `sentencepiece==0.2.0` works for e.g. that `slovlo-v1` model.

- Tom Aarsen

### DIFF
--- a/docker_images/sentence_transformers/requirements.txt
+++ b/docker_images/sentence_transformers/requirements.txt
@@ -1,6 +1,7 @@
 starlette==0.27.0
 api-inference-community==0.0.32
 sentence-transformers==3.0.1
+sentencepiece==0.2.0
 transformers==4.41.1
 tokenizers==0.19.1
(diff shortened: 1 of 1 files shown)


tomaarsen:
The same issue exists for some Camembert models, e.g.: https://huggingface.co/Photon-BR/sentence-camembert-large

- Tom Aarsen

osanseviero (approved):
Makes sense!

PR #439, opened 2024-07-24
docker_images: diffusers, do not start if pipeline fails to load

oOraph:


SBrandeis (approved):
Awesome thank you !!

oOraph:
closing pr: seen with @Narsil if we do so, the deployment will crashloop and the model will be loading indefinetely on widget side. Thus, people won't have the chance to get the failure trace for troubleshooting. I need to think about some other solution

PR #443, opened 2024-08-08
Possibility to override dtype trhough env var

oOraph:


apolinario (approved):
LGTM

apolinario:
Actually one more thing, we can update the diffusers to install `0.30.0` instead of from the commit I installed given it released yesterday

PR #447, opened 2024-08-27
Discard a request if the caller already left

oOraph:
No need to perform some inference that won't be fetched anyway


Narsil (approved):
LGTM

apolinario (approved):

PR #457, opened 2024-09-30
Reclassify model loading error as server errors

oOraph:


api_inference_community/routes.py
```diff
@@ -119,8 +119,6 @@ async def pipeline_route(request: Request) -> Response:
                     f'{error["msg"]}: received `{error["input"]}` in `parameters`'
                 )
         return JSONResponse({"error": errors}, status_code=400)
-    except (EnvironmentError, ValueError) as e:
-        return JSONResponse({"error": str(e)}, status_code=400)
```
SBrandeis, line 123: From a quick glance at transformers code, it seems that `ValueError` is mostly raised when the user passes an incorrect inference argument (which should be treated as HTTP 400)


.... buuuuut it's also seems to be used to express inference errors that the API user has no control over (such as an invalid model config), example:

https://github.com/huggingface/transformers/blob/4d5b45870411053c9c72d24a8e1052e00fe62ad6/src/transformers/pipelines/base.py#L268-L269


would it make sense to differentiate those two cases in transformers code base? cc @ArthurZucker

oOraph: good point. Side note: here we should be ok as there are two steps, model loading (get_pipeline) vs inference itself (__call__), in the first case, we can fairly consider ValueErrors as 500 and in the second we let them as 400 wdyt ?

Narsil (approved):
Still I don't feel it's a right move but I won't block anything over this.

PR #469, opened 2025-02-07
latent-to-image

hlky:
Context: https://huggingface.slack.com/archives/C065E480NN9/p1738859836468269 (internal)


api_inference_community/validation.py
```diff
@@ -196,8 +198,12 @@ def check_inputs(inputs, tag):
 IMAGE_OUTPUTS = {
     "image-to-image",
     "text-to-image",
+    "latent-to-image",
```
apolinario, line 201: Not sure about the new tag, as it maps to the existing `tasks` and I think we probably don't want to create this as an official task while this is experimental. But maybe this is harmless to exist only in the inference API

hlky: Was added for

https://github.com/huggingface/api-inference-community/blob/329d9bb61a463a9139f318e89619e4ed25fc9080/api_inference_community/routes.py#L229-L241

Seems to be the only usage of `IMAGE_OUTPUTS`

We also need the task in `KNOWN_TASKS` (via `TENSOR_INPUTS`)

https://github.com/huggingface/api-inference-community/blob/329d9bb61a463a9139f318e89619e4ed25fc9080/api_inference_community/routes.py#L96-L102

Should be harmless for this to exist but I agree we don't want it to be an official task while this is experimental.

.gitignore
```diff
@@ -9,3 +9,5 @@ pretrained_checkpoints
 *.egg-info/
 build/
 dist/
+
+.venv/
```
Narsil, line 13:
```suggestion
```

This is up to each developper. You need to use a global gitignore (otherwise we end up with an infinite list of every possible setup)

api_inference_community/routes.py
```diff
@@ -128,7 +129,9 @@ async def pipeline_route(request: Request) -> Response:
         return JSONResponse({"error": str(e)}, status_code=500)
 
     try:
-        inputs, params = normalize_payload(payload, task, sampling_rate=sampling_rate)
+        inputs, params = normalize_payload(
+            payload, task, sampling_rate=sampling_rate, headers=headers
```
Narsil, line 133: Probably not a good place to put the tensor shape in the headers. Why not just put it in the payload ?

I know base64 isn't perfect, but I'd rather go multipart upload than do something like this

api_inference_community/validation.py
Narsil: We can maybe find ways to "override" another task for the time being  (without involving another task)

api_inference_community/validation.py
```diff
@@ -407,3 +415,23 @@ def normalize_payload_nlp(bpayload: bytes, task: str) -> Tuple[Any, Dict]:
...
+        "float16": torch.float16,
+        "float32": torch.float32,
+        "bfloat16": torch.bfloat16,
+    }
+
+    shape = json.loads(headers.get("shape"))
+    dtype = DTYPE_MAP.get(headers.get("dtype"))
+    arr = torch.frombuffer(bytearray(bpayload), dtype=dtype).reshape(shape)
+    if sys.byteorder == "big":
+        arr = torch.from_numpy(arr.numpy().byteswap(inplace=False))
```
Narsil, line 433: We can definitely assume byteorder is little.

api_inference_community/routes.py
Narsil: Also changing this requires a release of the `api-inference-community` package which I think will slow the whole process down (because I will be way more picky about what gets actually merged since it's published as a package therefore harder to take down).

Can we start with base64 and change afterwards for more optimal ?

hlky: Changed to use base64. 

<details><summary>Minimal example</summary>
<p>

```python
import io
import base64
import requests
import torch
from PIL import Image
from safetensors.torch import _tobytes
inputs = torch.randn([1, 4, 64, 64], generator=torch.Generator().manual_seed(0))
shape = list(inputs.shape)
dtype = str(inputs.dtype).split(".")[-1]
tensor_data = base64.b64encode(_tobytes(inputs, "inputs")).decode("utf-8")
parameters = {"shape": shape, "dtype": dtype}
data = {"inputs": tensor_data, "parameters": parameters}

response = requests.post("http://127.0.0.1:8000/", json=data)
Image.open(io.BytesIO(response.content))
```

image


</p>
</details> 

<details><summary>Generation example</summary>
<p>

Using `Lykon/dreamshaper-8` finetune here for generation, can use `stable-diffusion-v1-5/stable-diffusion-v1-5` for VAE.

```python
import io
import base64
import requests
import torch
from PIL import Image
from safetensors.torch import _tobytes

from diffusers import DiffusionPipeline

pipeline = DiffusionPipeline.from_pretrained(
    "Lykon/dreamshaper-8",
    safety_checker=None,
    torch_dtype=torch.float16,
    variant="fp16",
).to("cuda")

prompt = "a cute cat sitting beside the sea beach."
inputs = pipeline(
    prompt=prompt,
    negative_prompt="bad quality, worse quality, degenerate quality",
    output_type="latent",
).images
shape = list(inputs.shape)
dtype = str(inputs.dtype).split(".")[-1]
tensor_data = base64.b64encode(_tobytes(inputs, "inputs")).decode("utf-8")
parameters = {"shape": shape, "dtype": dtype}
data = {"inputs": tensor_data, "parameters": parameters}

response = requests.post("http://127.0.0.1:8000/", json=data)
Image.open(io.BytesIO(response.content))
```

image



</p>
</details>

Narsil (approved):

